PR #11, opened 2014-08-11
Delete version attribute

mvano:
As discussed on public-webapps:
http://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0223.html


efullea:
already covered by PR #10

Issue #4, opened 2014-05-12
unregister should return Promise<PushRegistration>


mvano:
The WebIDL signature should become:
Promise&lt;PushRegistration&gt; unregister ();

This way, the page does not need to store the registration on the client. When it unregisters, it can take the registrationId from the PushRegistration and send it to the app server. The app server can then drop it from storage and stop sending messages to it.

mvano:
Maybe it should return only the registrationId, making the signature cleaner:
Promise&lt;DOMString&gt; unregister ();

johnmellor:
Hmm, but pushRegistrationIds might not be globally unique, instead they might only be unique for a given pushEndpoint. So I guess web apps need to send both the pushEndpoint and the pushRegistrationId to their server?

mvano:
Agreed with John, we should provide both.

efullea:
Agreed that it is the pair of pushRegistrationId and pushEndpoint what is globally unique. About how both parameters are provided to the server it is a matter of the app server -> push server interface. (e.g. pushEndpoint may be the URL where to POST a notification, and pushRegistrationId a parameter of the request). 

Comment addressed in a new commit (https://github.com/efullea/push-api/commit/efe64099c76c7c7bc8775ba08f4fde8f66586b11) to #14

Issue #5, opened 2014-07-03
isRegistered should return the endpoint itself rather than a boolean


prudentbot:
The way the current specification is defined, a developer would call isRegistered and, if the application is already registered, they then call register to obtain the endpoint.  The problem I see is that this results in a redundant query to local storage, one to check if the record exists (which is thrown away), and then, in most cases, one to obtain the record.  

If we made it so that isRegistered returned the result of that query rather than true/false, it would still perform the service of checking to see if the endpoint already exists, but then it would save the developer from having to call register() regardless of the result.  This has the added bonus of making the call to register() be more accurately descriptive, since it would only be used if you weren't already registered, and it would be used to perform the act of registration.

I suppose this implementation would also imply a name change from isRegistered to getRegistration or something.

PR #12, opened 2014-08-13
Single registration per webapp

efullea:
This PR fixes Issues #4 and #5


mvano:
Looks good to me. This is also in response to the discussion in:
http://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0223.html

Issue #2, opened 2014-05-05
Move pushRegistrationManager to ServiceWorker instance?


efullea:
Issue originally posted by @johnmellor at
https://github.com/telefonicaid/WebAPISpecs/issues/30

See issue and discussion there, but please post new comments here.

johnmellor:
See discussion [on public-webapps](http://comments.gmane.org/gmane.org.w3c.appformats/30135) and slightlyoff/ServiceWorker/issues/365.

Jake's suggestion (copied below) seems a very clean way of doing this, resolving the [earlier concerns](https://github.com/telefonicaid/WebAPISpecs/issues/30#issuecomment-41580520):

``` js
navigator.serviceWorker.ready.then(function(reg) {
  reg.push.register(...);
});
```

See also similar background sync issue: slightlyoff/BackgroundSync/issues/6

mvano:
Just updating John's code sample to the latest syntax:

``` idl
navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
  serviceWorkerRegistration.pushRegistrationManager.register().then(...);
});
```

efullea:
Closed by #13

PR #13, opened 2014-08-13
Push manager. Fixes issue #2

efullea:
Changes:
-Move pushRegistrationManager to ServiceWorkerRegistration (Fixes issue #2)
-PushManager --> PushRegistrationManager
-Adding PushRegistrationManager.hasPermission

### DIFF
--- a/index.html
+++ b/index.html
@@ -479,45 +479,24 @@ navigator.serviceWorker.ready.then(function() {
     </section>
     <section>
       <h2>
-        <a>Navigator</a> Interface
+        <a>ServiceWorkerRegistration</a> Interface
       </h2>
-      <dl title="partial interface Navigator" class="idl">
+      <dl title="partial interface ServiceWorkerRegistration" class="idl">
         <dt>
-          readonly attribute PushManager push
+          readonly attribute PushRegistrationManager pushRegistrationManager
         </dt>
         <dd>
           The object that exposes the push service management interface.
         </dd>
       </dl>
     </section>
-    <p class="note">
-      TAG issue: navigator.push as an object of PushManager type seems
-      misleading per se and is inconsistent with Array.prototype.push method.
-      Suggestion: navigator.pushRegistrationManager.
-    </p>
+
     <section>
       <h2>
-        <a>PushManager</a> Interface
+        <a>PushRegistrationManager</a> Interface
       </h2>
-      <p class="note">
-        TAG issue: PushManager is limited to dealing with registrations.
-        Suggestion: PushRegistrationManager.
-      </p>
-      <p class="note">
-        TAG issue: PushManager.registrations method seems inconsistent.
-        Suggestion: PushManager.getRegistrations.
-      </p>
-      <p class="note">
-        TAG issue: PushManager.register provides two heterogeneous functions:
-        asking use permission, registering push notifications. This mixing
-        leads to odd user agent behavior, i.e. push service failure forces
-        webapp to ask permission again. Suggestion: split the functionality and
-        provide separate methods to ask user permission (check if permission
-        has already being granted) and to register push notifications. Provide
-        different sets of errors for those purposes.
-      </p>
       <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -802,6 +790,36 @@
       </li>
    </ol>
      
+  <p> The <dfn><code>hasPermission</code></dfn> method when invoked MUST run the
+  following steps:
+    <ol>
+      <li>Let <var>promise</var> be a new <a><code>Promise</code></a> object and
```
beverloo: @mvano is changing the steps to not explicitly mention |resolver| in https://github.com/w3c/push-api/pull/23. He should probably rebase his patch on top of this.

index.html
```diff
@@ -1030,7 +1048,31 @@
...
+              permission to use Push API.
+          </dd>
+
+          <dt>denied</dt>
+          <dd>
+              The <a class="internalDFN" href="#dfn-webapp">webapp</a> has been
+              denied permission to use Push API.
+          </dd>
+
+          <dt>needToAsk</dt>
```
beverloo: The Web Notification API uses "default" for this, Mounir proposed a Permissions API to DAP today that uses "prompt", you're introducing "needToAsk" :-).

Could we settle on a single keyword? While I'm a fan of "prompt", "default" is already out there.

@mounirlamouri

index.html
```diff
@@ -802,6 +790,36 @@
...
+      algorithm with no arguments and terminate these steps.
+      </li>
+      <li>When the request has been completed invoke <em>resolver</em>'s <a class="internalDFN"
+          href="#dfn-fulfill-algorithm">fulfill algorithm</a> algorithm with
+          the <a><code>PushPermissionStatus</code></a> as the <code>value</code> argument.
+        </ol>
+      </li>
+   </ol>
+
+  <p> Permission to use the Push Service can be persistent, that is, it does not
```
beverloo: nit: s/Push Service/push service/. We don't capitalize it elsewhere.

index.html
```diff
@@ -656,6 +637,13 @@
 		notified about the result of the operation. 
     </dd>
     
+    <dt> Promise&lt;PushPermissionStatus&gt; hasPermission ()</dt>
+    <dd>This method allows a <a class="internalDFN"
+    href="#dfn-webapp">webapp</a> to check whether user has granted permission
+    to use push service. It returns a <a><code>Promise</code></a> that will
+    allow the caller to be notified about the result of the operation.
```
beverloo: ".. that will notify the caller about the result of the operation"? The proposed text makes it sound like the caller MAY want to do this, while the method is a no-op unless the result gets used.

beverloo:
Firstly, thank you for waiting!! :-)

While I don't remember my exact concern, one thing I do care about is availability of the PushRegistrationManager interface from within Service Workers. A possible scenario would be unregistering the active Push registration when the server sends a message that it's no longer necessary.

The ServiceWorkerRegistration interface is explicitly only exposed on Window, which means that it will no longer be able to use Push functionality (aside from removing messages) from within a worker. There is an open issue for this in the Service Workers repository:

https://github.com/slightlyoff/ServiceWorker/issues/421

It is noteworthy that @nikhilm points out in comment 12 that he would like this done after SW v1, and that Gecko won't expose push to SWs in the first place. We do want to expose this interface within workers.

index.html
efullea: 'prompt' looks more self-explanatory to me than 'default', so I would go for it.

mounirlamouri: I think _default_ was a safe word for _prompt_ in Notifications.

index.html
mounirlamouri: +1. This prose is actually missing requirement wording. It is better to explicitly say what what the UA should do and whether it is MUST or SHOULD (I assume that those are the options here).

efullea: Other methods seem to have the same wording. Let's deal with it separately.

PR #20, opened 2014-08-27
Clean up the JavaScript example

mvano:
This brings the example much closer to current discussions. This does not yet take into account the idea of moving pushRegistrationManager to ServiceWorkerRegistration as I think Peter had some questions about that.

This fixes issue #17.

### DIFF
--- a/index.html
+++ b/index.html
@@ -343,9 +343,13 @@
       </p>
       <p>
         The following code and diagram illustrate a hypothetical use of the
-        push API:
+        push API.
       </p>
-      <pre class="example highlight">
+      <section class="informative">
+        <h2>
+          Example
+        </h2>
+        <pre class="example highlight">
 // https://example.com/serviceworker.js
 this.onpush = function(event) {
   // During development we can log the data for easier inspection.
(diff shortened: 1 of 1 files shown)


efullea:
Bryan (@blsaws) was the author of the example. Probably he wants to have a look.

mvano:
Afaik Bryan actually copied that example from a discussion, possibly with some modifications. Either way it was a rough first cut and both Service Workers and Push have changed since then. I think since he has not commented on this pull request, anyone should feel free to review it.

index.html
```diff
@@ -360,102 +360,29 @@
...
+}
+
+// https://example.com/webapp.js
+navigator.serviceWorker.ready.then(function() {
+  navigator.pushRegistrationManager.register().then(
+    function(registration) {
+      console.log(registration.pushRegistrationId);
+      console.log(registration.pushEndpoint);
+      // We now have the registration details that the application server needs
+      // to start pushing messages. From here we can send them to th  server
```
beverloo: nit: s/th /the/.

index.html
```diff
@@ -360,102 +360,29 @@
...
+      // using an asynchronous XMLHttpRequest.
```
beverloo: Perhaps clarify the comment a little bit to (a) indicate where the application server needs to send the messages to, and (b) prevent making good practices sound like requirements?

"The push registration details needed by the application server to push messages to the push service are now available, and can be send to the application service using, for example, an XMLHttpRequest."

index.html
```diff
@@ -360,102 +360,29 @@
...
+    }, function(e) {
+      console.log(e);
```
beverloo: Since this is an example, please annotate why it would be good practice for a user to log |e| here.

mvano:
Thanks Peter! Please take another look.

index.html
```diff
@@ -343,102 +343,42 @@
...
-    // offer the "local" push server for this registration.
-    sender: {
-     apn: apnSenderId,
-     gcm: gcmSenderId,
-     // ...
+// https://example.com/webapp.js
+navigator.serviceWorker.ready.then(function() {
+  navigator.pushRegistrationManager.register().then(
+    function(registration) {
+      // During development we can log the registration details for easier
```
beverloo: micro nit: you repeat "During development" trice. Perhaps removing the comments on lines 355 and 365 might make sense?

beverloo:
Looks good to me. Thanks Michael!

mvano:
Thanks Peter. I don't feel strongly about those comments so I removed them.

@efullea I think this is ready to merge. Could you take a look? Thanks!

PR #21, opened 2014-08-28
Clean up ReSpec config

mvano:
- Delete noisy comments and point to ReSpec documentation
- Update url to current ED
- Delete prettyprint, inlineCSS, and noIDLIn as they don't do anything
- Add subjectPrefix
- Delete double nested example markup


efullea:
LGTM

PR #22, opened 2014-08-28
Clean up terminology section

mvano:
- Update Promise reference to ECMASCRIPT and put it next to other dependencies
- Less verbose


efullea:
LGTM

PR #23, opened 2014-08-28
Cleaner steps for register, unregister, getRegistration

mvano:
- Less verbose, no need to refer to resolver.
- Use DOMException, not DOMError.
- It's not ok to reject a promise without arguments, always provide a DOMException.
- Resolve unregister promise with PushRegistration.

For more background details see the Promises Guide:
https://github.com/w3ctag/promises-guide

This fixes issue #4.

### DIFF
--- a/index.html
+++ b/index.html
@@ -643,7 +643,7 @@
             href="#dfn-webapp">webapp</a> to receive <a class="internalDFN"
             href="#dfn-push-message">push messages</a>, unless a prearranged
             trust relationship applies or the user has already granted or
-            denied permission explicitly to this <a class="internalDFN" href=
+            denied permission explicitly for this <a class="internalDFN" href=
             "#dfn-webapp">webapp</a>.
           </li>
           <li>If not granted, reject <var>promise</var> with a <a href=
@@ -666,7 +666,7 @@
               "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>"
               and terminate these steps.
               </li>
-              <li>When the request has been completed resolve
+              <li>When the request has been completed, resolve
               <var>promise</var> with a <a><code>PushRegistration</code></a>
               providing the details of the retrieved <a class="internalDFN"
               href="#dfn-push-registration">push registration</a>.
@@ -683,12 +683,16 @@
           "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>"
           and terminate these steps.
           </li>
-          <li>When the request has been completed resolve <var>promise</var>
+          <li>When the request has been completed, resolve <var>promise</var>
           with a <a><code>PushRegistration</code></a> providing the details of
           the new <a class="internalDFN" href="#dfn-push-registration">push
           registration</a>.
           </li>
         </ol>
+        <p class="note">
+          <a href=
+          "https://www.w3.org/Bugs/Public/show_bug.cgi?id=23033"><code>PermissionDeniedError</code></a> has not yet been defined in a specification, this document currently links to the bug for defining it.
+        </p>
         <p>
           The <dfn><code>unregister</code></dfn> method when invoked MUST run
           the following steps:
@@ -717,10 +721,10 @@
           "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>"
           and terminate these steps.
           </li>
-          <li>When the request has been completed resolve <var>promise</var>
+          <li>When the request has been completed, resolve <var>promise</var>
           with a <a><code>PushRegistration</code></a> providing the details of
-          the unregistered <a class="internalDFN" href=
-          "#dfn-push-registration">push registration</a>.
+          the <a class="internalDFN" href="#dfn-push-registration">push
+          registration</a> which has been unregistered.
           </li>
         </ol>
         <p>
@@ -750,7 +754,7 @@
           "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>"
           and terminate these steps.
           </li>
-          <li>When the request has been completed resolve <var>promise</var>
+          <li>When the request has been completed, resolve <var>promise</var>
           with a <a><code>PushRegistration</code></a> providing the details of
           the retrieved <a class="internalDFN" href=
           "#dfn-push-registration">push registration</a>.
--- a/index.html
+++ b/index.html
@@ -598,7 +598,7 @@
     notified about the result of the operation. 
     </dd>
 
-    <dt> Promise&lt;DOMString&gt; unregister ()</dt>
+    <dt> Promise&lt;PushRegistration&gt; unregister ()</dt>
     <dd>This method allows a <a class="internalDFN"
     href="#dfn-webapp">webapp</a> to unregister. It
     returns a <a><code>Promise</code></a> that will allow the caller to be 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -663,121 +663,80 @@
...
-          href="#dfn-reject-algorithm">reject algorithm</a> with <em>error</em>
-          as the <code>value</code> argument.
-        </ol>
-      </li>
-      <li>Make a request to the system to create a new <a class="internalDFN"
+      href="#dfn-webapp">webapp</a>.</li>
+    <li>If not granted then reject <var>promise</var> with a <a href=
+      "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+      whose name is "<a href=
+      "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>" and
```
nikhilm: Can we use PermissionDeniedError - https://www.w3.org/Bugs/Public/show_bug.cgi?id=23033?

index.html
```diff
@@ -663,121 +663,80 @@
...
-      </li>
-      <li>Return <var>promise</var> and continue the following steps
-      asynchronously.
-      </li>
-      <li>Make a request to the system to deactivate the <a class="internalDFN"
+      requesting <a class="internalDFN" href="#dfn-webapp">webapp</a>.</li>
+    <li>If there is an error then reject <var>promise</var>
+      with a <a href=
+      "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+      of the most appropriate type and terminate these steps.</li>
```
nikhilm: This should be specced. We don't want different UAs returning different errors if their underlying systems fail. None of the DOMException errors are very appropriate though for the 'we screwed up' case :(

index.html
```diff
@@ -663,121 +663,80 @@
...
-      </li>
-      <li>Make a request to the system to create a new <a class="internalDFN"
+      href="#dfn-webapp">webapp</a>.</li>
+    <li>If not granted then reject <var>promise</var> with a <a href=
+      "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+      whose name is "<a href=
+      "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>" and
+      terminate these steps.</li>
+    <li>Make a request to the system to create a new <a class="internalDFN"
       href="#dfn-push-registration">push registration</a> for the
```
nikhilm: also, reuse the existing registration if any.

index.html
```diff
@@ -663,121 +663,80 @@
...
+    <li>Return <var>promise</var> and continue the following steps
+      asynchronously.</li>
+    <li>Retrieve the <a class="internalDFN"
+      href="#dfn-push-registration">push registration</a> associated with the
+      requesting <a class="internalDFN" href="#dfn-webapp">webapp</a>.</li>
+    <li>If there is an error or the <a class="internalDFN"
+      href="#dfn-webapp">webapp</a> is not registered then reject
+      <var>promise</var> with a <a href=
+      "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+      of the most appropriate type and terminate these steps.</li>
```
nikhilm: NotFoundError?

index.html
```diff
@@ -663,121 +663,80 @@
...
-      </li>
-      <li>When the request has been completed invoke <em>resolver</em>'s <a
-      class="internalDFN" href="#dfn-fulfill-algorithm">fulfill algorithm</a>
-      with a <code>DOMString</code> set to the value of the
+      class="internalDFN" href="#dfn-webapp">webapp</a>.</li>
+    <li>If there is an error then reject <var>promise</var> with a <a href=
+      "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+      of the most appropriate type and terminate these steps.</li>
+    <li>When the request has been completed resolve <var>promise</var> with a
+      <code>DOMString</code> set to the value of the
```
nikhilm: Could we have this also resolve to a PushRegistration?

index.html

index.html

index.html
mvano: I'm fine with either of these approaches:
1. The spec says to use a DOMException. The implementation can then use different types as needed e.g. NetworkError or NotReadableError. I think there is value in permitting an implementation to use informative errors this way.
2. The spec says to use the most appropriate known error. I'd suggest using AbortError. The implementation has no choice of error type, but can still provide more specific information to the author using the message attribute. This is not to be coded against (though this will inevitably happen).

index.html
mvano: Done, this resolves issue #4.

index.html

mvano:
All done, thanks Nikhil!

mvano:
Any further comments? Otherwise I'd like to get this merged. I think this is an improvement, and I'm happy to iterate further in subsequent (but probably smaller) patches.

index.html
```diff
@@ -661,75 +661,71 @@
...
-          "internalDFN" href="#dfn-webapp">webapp</a> to activate the reception
-          of <a class="internalDFN" href="#dfn-push-message">push messages</a>,
-          unless a prearranged trust relationship applies or the user has
-          already granted or denied permission explicitly to this <a class=
-          "internalDFN" href="#dfn-webapp">webapp</a>.
+          <li>Ask the user whether they allow the <a class="internalDFN"
+            href="#dfn-webapp">webapp</a> to receive <a class="internalDFN"
+            href="#dfn-push-message">push messages</a>, unless a prearranged
+            trust relationship applies or the user has already granted or
+            denied permission explicitly to this <a class="internalDFN" href=
```
beverloo: nit: s/to/for/.

index.html
```diff
@@ -661,75 +661,71 @@
...
+            "#dfn-webapp">webapp</a>.
           </li>
-          <li>If not granted, run the following sub-steps and terminate these
-          steps:
+          <li>If not granted, reject <var>promise</var> with a <a href=
+          "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+          whose name is "<a href=
+          "https://www.w3.org/Bugs/Public/show_bug.cgi?id=23033"><code>PermissionDeniedError</code></a>"
```
beverloo, line 652: We shouldn't link to an unresolved bug in normative text, without explicitly noting this as an open issue.

index.html
```diff
@@ -661,75 +661,71 @@
...
+          and terminate these steps.
+          </li>
+          <li>If the <a class="internalDFN" href="#dfn-webapp">webapp</a> is
+          already registered, run the following substeps:
             <ol>
```
beverloo, line 657: I'm not a fan of the duplication this has with the steps for retrieving the current registration. Is there precedence for phrasing this like: "Run the steps defined in the [retrieve registration] algorithm, and resolve |promise| with the returned promise."?

index.html
```diff
@@ -661,75 +661,71 @@
...
-              <code>PushRegistration</code> object providing the details of the
-              <a class="internalDFN" href="#dfn-push-registration">push
-              registration</a>.
-              </li>
-              <li>Invoke <em>resolver</em>'s <a class="internalDFN" href=
-              "#dfn-fulfill-algorithm">fulfill algorithm</a> with
-              <em>pushRegistration</em> as the <code>value</code> argument.
-              </li>
-            </ol>
+          <li>When the request has been completed resolve <var>promise</var>
```
beverloo: nit: s/completed/completed,/

index.html
```diff
@@ -737,26 +733,33 @@
...
-          "internalDFN" href="#dfn-push-registration">push registration</a> of
-          the requesting <a class="internalDFN" href="#dfn-webapp">webapp</a>.
+          "internalDFN" href="#dfn-push-registration">push registration</a>
+          associated with the <a class="internalDFN" href=
+          "#dfn-webapp">webapp</a>.
           </li>
-          <li>If there is an error invoke <em>resolver</em>'s <a class=
-          "internalDFN" href="#dfn-reject-algorithm">reject algorithm</a> with
-          no arguments and terminate these steps.
+          <li>If there is an error, reject <var>promise</var> with a <a href=
```
beverloo, line 720: Perhaps we could add a non-normative note indicating that even if the UA was unable to withdraw the registration with the push service, it should ensure that no further messages are delivered to the SW? Right now it's undefined what happens when an AbortError gets thrown.

index.html
```diff
@@ -737,26 +733,33 @@
...
+          "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>"
+          and terminate these steps.
           </li>
-          <li>When the request has been completed invoke <em>resolver</em>'s
-          <a class="internalDFN" href="#dfn-fulfill-algorithm">fulfill
-          algorithm</a> with a <code>DOMString</code> set to the value of the
-          <code>pushRegistrationId</code> attribute of the unregistered
-          <a class="internalDFN" href="#dfn-push-registration">push
-          registration</a> as the <code>value</code> argument.
+          <li>When the request has been completed resolve <var>promise</var>
```
beverloo: nit: s/completed/completed,/

index.html
```diff
@@ -737,26 +733,33 @@
...
+          with a <a><code>PushRegistration</code></a> providing the details of
+          the unregistered <a class="internalDFN" href=
```
beverloo: ".. the details of the push registration which has been unregistered"? It clarifies slightly better that the information contained in the registration has been invalidated.

index.html
```diff
@@ -764,32 +767,32 @@
...
           </li>
-          <li>If there is an error or the <a class="internalDFN" href=
-          "#dfn-webapp">webapp</a> is not registered invoke <em>resolver</em>'s
-          <a class="internalDFN" href="#dfn-reject-algorithm">reject
-          algorithm</a> algorithm with no arguments and terminate these steps.
+          <li>If there is an error, reject <var>promise</var> with a <a href=
+          "http://dom.spec.whatwg.org/#domexception"><code>DOMException</code></a>
+          whose name is "<a href=
+          "http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>"
+          and terminate these steps.
```
beverloo, line 757: What is the benefit of passing an AbortError if there's only one possible path leading to rejection ("an error")? How would the developer use this information? Do you envision additional rejection paths in the future?

Of course, this would be fine if it matches what other Promise-using specifications do.

index.html
```diff
@@ -764,32 +767,32 @@
...
-              <li>Let <em>pushRegistration</em> be the
-              <code>pushRegistration</code> object associated to the requesting
-              <a class="internalDFN" href="#dfn-webapp">webapp</a>.
-              </li>
-              <li>Invoke <em>resolver</em>'s <a class="internalDFN" href=
-              "#dfn-fulfill-algorithm">fulfill algorithm</a> algorithm with
-              <em>pushRegistration</em> as the <code>value</code> argument.
-              </li>
-            </ol>
+          <li>When the request has been completed resolve <var>promise</var>
```
beverloo: nit: s/completed/completed,/

index.html
mvano, line 652: Done.

index.html
mvano, line 657: I think it's fair to link to larger algorithms to avoid repeating them, but this would link to just the last 3 steps of getRegistration. Adding that indirection in two places does not seem helpful for 3 steps.

index.html

index.html
mvano, line 757: The promise can be rejected with either a NotFoundError or an AbortError, so there are two paths. Additionally, the UA has the option to provide details in error.message that cannot be specified in this document, but would still be useful for authors when logged to the console.

index.html
mvano, line 720: The spec does not know why the AbortError was thrown. It could be because the UA failed to modify storage, in which case perhaps it can also not remember that unregister was called. I think your suggestion is possible, but I'd rather avoid having maybe-maybe-not non-normative notes. You're welcome to file an issue for this so we can further explore the ramifications of black-box system errors. The behavior should then also be consistent across the spec.

mvano:
Thanks Peter! I think this is ready for another look.

index.html
```diff
@@ -622,102 +622,111 @@
...
-          of <a class="internalDFN" href="#dfn-push-message">push messages</a>,
-          unless a prearranged trust relationship applies or the user has
-          already granted or denied permission explicitly to this <a class=
-          "internalDFN" href="#dfn-webapp">webapp</a>.
+          <li>Ask the user whether they allow the <a class="internalDFN"
+            href="#dfn-webapp">webapp</a> to receive <a class="internalDFN"
+            href="#dfn-push-message">push messages</a>, unless a prearranged
+            trust relationship applies or the user has already granted or
+            denied permission explicitly for this <a class="internalDFN" href=
+            "#dfn-webapp">webapp</a>.
```
beverloo, line 647: nit: this indentation is inconsistent with the rest of the file.

beverloo:
SGTM

index.html
mvano, line 647: Agreed, but that's what tidy makes of it. I don't think we should fight the tool.

beverloo, line 647: If tidy is introducing inconsistent indentation then that is something which should be fixed. Either by reformatting the rest of the spec, or changing tidy's settings.

mvano:
@efullea I think this is ready to be merged. Could you take a look please? Thanks!

index.html
mvano, line 647: Fixing tidy seems like a separate change. I filed issue #39 for this.

Issue #32, opened 2014-09-01
Remove pushRegistrationId attribute from push event


mvano:
Currently there's both `data` and `pushRegistrationId` on a push event. I see no real use case for the id. Is this a leftover from when the api supported multiple registrations / channels? Let's remove it.

If the author really wants to have the id it can be read from `getRegistration` and we can always add it back if we find a strong use case.

PR #34, opened 2014-09-02
Remove pushRegistrationId from PushMessage. Fixes issue #32.

mvano:
- Remove pushRegistrationId from PushMessage
- While removing the step for setting pushRegistrationId I noticed the step for setting data was missing, so I added that while I was touching that section anyway.


beverloo:
LGTM

Issue #31, opened 2014-09-01
Rename #pushEndpoint to #endpoint


mvano:
Currently `PushRegistration` has an attribute named `pushEndpoint`. This seems overly verbose to me. Suggestion: rename it `endpoint`.

Issue #30, opened 2014-09-01
Rename #pushRegistrationId to #registrationId


mvano:
Currently `PushRegistration`, `PushMessage`, and `PushRegisterMessage` have an attribute named `pushRegistrationId`. This seems overly verbose to me. Suggestion: rename it to `registrationId`.

PR #36, opened 2014-09-03
Rename pushEndpoint to endpoint. Fixes issue #31

mvano:


beverloo:
It seems inconsistent to rename pushEndpoint to endpoint, but not rename pushRegistrationId to registrationId.

beverloo:
Ah, that's issue #30. Why not do both in the same patch?

It would be great if we could get some more input on these issues.

mvano:
I'm splitting these into small changes where possible, to avoid merge conflicts and make reviewing easier.

Agreed that more input from more people and companies would be welcome. While we're working on Github I think visibility is a bit limited. However, the current state of the spec is inconsistent, and asking people to read it is likely to be wasteful. Once the spec is more or less up to date with the discussions and implementations I'd like to request more input from public-webapps.

beverloo:
LGTM, when both will be applied. The interface name (PushRegistration) itself is clear enough.

mvano:
Also renamed `pushRegistrationId` to `registrationId`.

mvano:
@efullea I think this is ready to be merged.

mvano:
Rebased. Should be good now @efullea

PR #37, opened 2014-09-04
Add viewport meta tag

mvano:


index.html
```diff
@@ -5,6 +5,7 @@
       Push API
     </title>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width">
```
beverloo: This seems to be rather uncommon. Several other specifications seem to achieve resolution independence by setting "max-width: 50em" on <body>.

If you do decide to go this way, you should also specify initial-scale=1.0 (per http://mobile-html.appspot.com/#14)

mvano: Added initial-scale=1.0. I'll see about further styling later.

beverloo:
SGTM

mvano:
@efullea I think this is ready to be merged. Could you take a look please? Thanks!

Issue #49, opened 2014-09-30
Abstract does not provide enough information about Push API


mvano:
The abstract has a few issues:
- It does not mention service workers though they are a critical dependency that shapes a large part of the API.
- It refers to a list of other specifications that are not normative. Most of them seem not very relevant to the reader, at least not when reading an abstract. Perhaps some of this can help clarify some details further down in the document.

PR #50, opened 2014-09-30
Abstract should mention service worker and drop non-normative references

mvano:
This addresses issue #49

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,32 +53,13 @@
   <body>
     <section id='abstract'>
       <p>
-        The <cite>Push API</cite> provides <a class="internalDFN" href=
-        "#dfn-webapp">webapps</a> with scripted access to server-sent messages,
-        for simplicity referred to here as <a class="internalDFN" href=
-        "#dfn-push-message">push messages</a>, as delivered by <a class=
-        "internalDFN" href="#dfn-push-service">push services</a>. A <a class=
-        "internalDFN" href="#dfn-push-service">push service</a> allows a
-        <a class="internalDFN" href="#dfn-webappserver">webapp server</a> to
-        send messages to a <a class="internalDFN" href=
-        "#dfn-webapp">webapp</a>, regardless of whether the <a class=
-        "internalDFN" href="#dfn-webapp">webapp</a> is actively running at the
-        time or not. The <a>push message</a> is delivered to a <a class=
-        "internalDFN" href="#dfn-service-worker">Service Worker</a> as a push
-        event. The push event optionally has a data payload. From the event
-        handler any available API may be called, for example to store the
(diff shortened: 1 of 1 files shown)


beverloo:
This is really hard to review because of the overhead caused by all the HTML annotations, so I wrote down the old and the new version in text:

Previous:

> This specification defines a "Push API" that provides [webapps] with scripted access to server-sent messages, for simplicity referred to here as [push messages], as delivered by [push services]. [Push services] are a way for [webapp servers] to send messages to [webapps], whether or not the [webapp] is active in a browser window.
> 
> [Push messages] may be delivered to the [user agent] via various messages, either via standardized protocols (e.g. Server-Sent Events, the GSM Short Message Service, SIP MESSAGE or OMA Push), or via [user agent] specific methods. The method to be used is selected by the [user agent], or it may allow the user to select one. The Push API is defined to promote compatibility with any delivery method.

New:

> The Push API provides [webapps] with scripted access to server-sent messages, for simplicity referred to here as [push messages], as delivered by [push services]. A [push service] allows a [webapp server] to send messages to a [webapp], regardless of whether the [webapp] is actively running at the time or not. The [push message] is delivered to a [Service worker] as a push event. The push event optionally has a data payload. From the event handler any available API may be called, for example to store the message data or to display a notification to the user.
> 
> [Push messages] may be delivered by the [push service] to the [user agent] via any method, this part of the system is not exposed as part of the Push API. The method to be used is selected by the [user agent], or it may allow the user to select one. The Push API is defined to promote compatibility with any delivery method.

Let me reply to a few of your changes.
- You changed "whether or not the webapps is active in a browser window" to "regardless of whether the webapp is actively running at the time or not". This causes some ambiguity over the definition of "is actively running" -- what does it mean? Perhaps something like "regardless of whether the webapp is currently active on the [user agent]" would be better.
- I think that the "The push message...may be called" bit in the abstract goes into too much detail. What about replacing it with "The [push message] will be delivered to a [Service Worker], which could then store the message's data or display a notification to the user." We may even want to consider mentioning "background" here.
- The entire second paragraph is verbosely saying that the protocol used to deliver push messages is considered out of scope for this specification. Could we rephrase it as "This specification is designed to promote compatibility with any delivery method for [push messages]."?
- In the IETF Web Push protocol we use "application server" instead of "webapp server". Aligning the two phrases would be good.

mvano:
Thanks Peter! All addressed. Please take another look.
I filed issue #53 for the app server terminology.

beverloo:
Thank you. I think this captures it well, but would like some other opinions on this as well.

@efullea @mounirlamouri

efullea:
LGTM

PR #51, opened 2014-09-30
Remove generated code

mvano:
It looks like at some time in the past, markup generated by respec was committed as source. Or maybe these are artifacts from before this document used respec.

This will be hard to review, but these are the changes:
- dfn tags don't need an id attribute
- a tags to internal references only need a title attribute (and never any others) if the text content does not match the dfn content as might happen with plurals
- rfc2119 keywords need no markup when they're uppercase

### DIFF
--- a/index.html
+++ b/index.html
@@ -54,26 +54,21 @@
   <body>
     <section id='abstract'>
       <p>
-        This specification defines a “Push API” that provides <a class=
-        "internalDFN" href="#dfn-webapp">webapps</a> with scripted access to
-        server-sent messages, for simplicity referred to here as <a class=
-        "internalDFN" href="#dfn-push-message">push messages</a>, as delivered
-        by <a class="internalDFN" href="#dfn-push-service">push services</a>.
-        <a class="internalDFN" href="#dfn-push-service">Push services</a> are a
-        way for <a class="internalDFN" href="#dfn-webappserver">webapp
-        servers</a> to send messages to <a class="internalDFN" href=
-        "#dfn-webapp">webapps</a>, whether or not the <a class="internalDFN"
-        href="#dfn-webapp">webapp</a> is active in a browser window.
+        This specification defines a “Push API” that provides <a title="webapp">webapps</a> with scripted access to
+        server-sent messages, for simplicity referred to here as <a title="push message">push messages</a>, as delivered
+        by <a title="push service">push services</a>.
+        <a title="push service">Push services</a> are a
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -198,63 +175,48 @@
...
+        <dfn><a href=
         "http://www.w3.org/TR/service-workers/#dnf-service-worker">Service
         Worker</a></dfn> is defined in [[!SERVICE-WORKERS]].
       </p>
       <p>
-        The term <dfn id="dfn-webapp">webapp</dfn> refers to a Web application,
-        i.e. an application implemented using Web technologies, and executing
-        within the context of a Web <a class="internalDFN" href=
-        "#dfn-user-agent">user agent</a>, e.g. a Web browser or other Web
+        The term <dfn>webapp</dfn> refers to a Web application, i.e. an
```
beverloo: Given that we need a title attribute on each plural (`<a title="webapp">webapps</a>`), it could be interesting to support some kind of `data-aliases` attribute for `<dfn>` accepting a space-separated list of aliases. Having `<dfn data-aliases="webapps">webapp</dfn>` once beats having title attributes for all of each plural.

Filed as https://github.com/w3c/respec/issues/344

+@darobin

index.html
```diff
@@ -262,31 +224,26 @@
...
-        <em title="must" class="rfc2119">must not</em> provide Push API access
-        to <a class="internalDFN" href="#dfn-webapp">webapps</a> without the
-        <a class="internalDFN" href="#dfn-express-permission">express
-        permission</a> of the user. <a class="internalDFN" href=
-        "#dfn-user-agent">User agents</a> <em title="must" class=
-        "rfc2119">must</em> acquire consent for permission through a user
-        interface for each call to the <code>register()</code> method, unless a
-        prearranged trust relationship applies.
+        <a title="user agent">User agents</a> MUST NOT provide Push API access
+        to <a title="webapp">webapps</a> without the <a>express permission</a>
```
beverloo: unrelated nit: "express permission"? Explicit permission? Would the "allowance" brain-fart actually be against the spec?

beverloo:
Yay! LGTM

Could consider moving to a higher line length limit too to increase readability even further.

index.html
mvano: Filed issue #52 for this.

martinthomson:
Re: line length:  http://baymard.com/blog/line-length-readability

mvano:
All that research about line lengths was not about indented html was it ;-)

PR #54, opened 2014-09-30
Fix respec warnings and a bad idl definition

mvano:


index.html
```diff
@@ -461,9 +461,9 @@
               is "<a href="http://dom.spec.whatwg.org/#aborterror"><code>AbortError</code></a>" and
               terminate these steps.
               </li>
-              <li>When the request has been completed, resolve <var>promise</var> with a <a><code>
-                PushRegistration</code></a> providing the details of the retrieved <a>push
-                registration</a>.
+              <li>When the request has been completed, resolve <var>promise</var> with a
+              <code>PushRegistration</code> providing the details of the retrieved <a>push
```
beverloo, line 465: Should lines 389 and 621 also use `<code>`? What is the difference between `<a><code>def</code></a>` and one without an anchor?

mvano, line 465: I filed issue #58 for making this consistent throughout. The current patch just clears respec warnings. `code` is just visual styling, `a` is different visual styling and linking.

PR #57, opened 2014-10-01
Set noLegacyStyle: true and adjust structure accordingly

mvano:
See http://www.w3.org/respec/ref.html#nolegacystyle for documentation.

Highlights:
- No auto-generated sections for attributes and methods, so less noise in the ToC and further down.
- The generated sections were sorted alphabetically so often differed from the WebIDL ordering, which is confusing. The hand coded markup is not re-ordered.

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,7 +46,9 @@
           // This is important for Rec-track documents, do not copy a patent URI
           // from a random document unless you know what you're doing. If in
           // doubt ask your friendly neighbourhood Team Contact.
-          wgPatentURI:  "http://www.w3.org/2004/01/pp-impl/42538/status"
+          wgPatentURI:  "http://www.w3.org/2004/01/pp-impl/42538/status",
+
+          noLegacyStyle: true
       };
     </script>
   </head>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -623,27 +620,30 @@
...
+      <p>
+        When getting the <code id="widl-PushRegistration-endpoint">endpoint</code> attribute, the
+        <a>user agent</a> MUST return the absolute URL exposed by the <a>push server</a> where the
+        <a>webapp server</a> can send <a title="push message">push messages</a> to this
+        <a>webapp</a>. The value of <code>endpoint</code> may be the same for multiple <a title=
+        "webapp">webapps</a> / <a>webapp</a> instances running on multiple devices.
+      </p>
+      <p>
+        When getting the <code id="widl-PushRegistration-registrationId">registrationId</code>
+        attribute the <a>user agent</a> MUST return a univocal identifier of this <a>push
```
beverloo: nit: comma after "attribute"

index.html
```diff
@@ -652,15 +652,16 @@
         <dt>
           readonly attribute DOMString? data
         </dt>
-        <dd>
-          MUST return the message data received by the <a>user agent</a> in the <a>push
-          message</a>, or null if no data was received.
-        </dd>
       </dl>
+      <p>
+        When getting the <code id="widl-PushMessage-data">data</code> attribute the <a>user
```
beverloo: nit: comma after "attribute"

index.html
```diff
@@ -675,11 +676,12 @@
         <dt>
           readonly attribute DOMString registrationId
         </dt>
-        <dd>
-          MUST return the identifier of the <a>push registration</a> to which this event is
-          related.
-        </dd>
       </dl>
+      <p>
+        When getting the <code id="widl-PushRegisterMessage-registrationId">registrationId</code>
```
beverloo, line 681: dito

index.html
```diff
@@ -786,22 +788,44 @@
...
-          The <a>webapp</a> has been denied permission to use Push API.
-        </dd>
         <dt>
           default
         </dt>
-        <dd>
-          The <a>webapp</a> needs to ask for permission in order to use Push API.
-        </dd>
       </dl>
+      <table class="simple">
```
beverloo, line 798: Well this is a bit sad. Technically it is a definition list, so `<dl>` would be more appropriate than a `<table>` element. Are you sure that this can't be done with a `<dl>`?

index.html

index.html

index.html
mvano, line 798: This renders the exact same thing as before. The `<dl>` now creates just the WebIDL snippet. The documentation table is what was previously generated by respec. Sadly, we don't have more granular control over what respec does and does not generate. Overall, I'm happy with this patch as it makes the document more readable, less cluttered. The underlying markup is less important.

Issue #33, opened 2014-09-01
Replace PushMessage interface with PushEvent


mvano:
The spec currently says to fire a push event which has a `PushMessage`. This means that reading data from it looks like this:

``` javascript
var data = event.pushmessage.data;
```

It seems cleaner to just define `PushEvent` like this:

``` idl
[Constructor(DOMString type, optional PushEventInit eventInitDict), Exposed=ServiceWorker]
interface PushEvent : Event {
  readonly attribute DOMString? data;
};

dictionary PushEventInit : EventInit {
  DOMString? data;
};
```

So the resulting javascript looks like this:

``` javascript
var data = event.data;
```

prudentbot:
Gecko is going to implement this solution for now, since it seems like it's most in keeping with current web platform conventions.

mvano:
@ndvr That's great to hear! I'd like to also draw your attention to the discussion of how to expose `waitUntil` on the event: https://github.com/slightlyoff/ServiceWorker/issues/448

Issue #18, opened 2014-08-13
PushRegisterMessage => PushError


efullea:
PushRegisterMessage => PushError with 'reason' attribute (enumeration). Now it covers at least 2 type of errors: 
1- new registration is needed (e.g. due to fatal error at server), 
2- "out of sync" , informational and does not imply registration has ended

mvano:
Can we phrase this so that it inherits from DOMException? The reason afaik should be a string. The type could be RegistrationMissingError or OutOfSyncError. I'm not sure what the formal pattern looks like exactly.

See: http://www.w3.org/TR/WebIDL/#idl-exceptions

mvano:
I think the enum approach would look as follows in WebIDL:

``` idl
[Constructor(DOMString type, optional PushErrorEventInit eventInitDict), Exposed=ServiceWorker]
interface PushErrorEvent : Event {
  readonly attribute PushErrorReason reason;
};

dictionary PushErrorEventInit : EventInit {
  PushErrorReason reason;
};

enum PushErrorReason {
  "registration-missing",
  "out-of-sync"
};
```

So the javascript would then look like:

``` javascript
this.onerror = function(event) {
  switch(event.reason) {
    case "registration-missing":
     ...
    break;
    case "out-of-sync":
     ...
    break;
  }
};
```

mvano:
The enum approach has a drawback: it is basically reinventing `DOMException` by replacing `name` or `code` with `reason`. I think we can reuse `DOMException` by using an instance of it for `PushErrorEvent`:

``` idl
[Constructor(DOMString type, optional PushErrorEventInit eventInitDict), Exposed=ServiceWorker]
interface PushErrorEvent : Event {
  readonly attribute DOMException error;
};

dictionary PushErrorEventInit : EventInit {
  DOMException error;
};
```

We can then add `PushOutOfSync` and `PushRegistrationLost` to the table of error names - http://www.w3.org/TR/dom/#error-names-0

So the javascript would then look like:

``` javascript
this.onerror = function(event) {
  switch(event.error.name) {
    case "PushRegistrationLost":
     ...
    break;
    case "PushOutOfSync":
     ...
    break;
  }
};
```

I think the DOMException approach has two drawbacks of its own:
- The `event.error.name` syntax is verbose and a bit weird
- It requires adding very Push-specific error names

annevk:
Maybe this can reuse http://www.whatwg.org/specs/web-apps/current-work/#the-errorevent-interface in some way?

[...]

PR #59, opened 2014-10-02
Clean up events and define relationship to service workers

mvano:
This relates to issues #33 and #18.
- Define event handlers `onpush` and `onpushregistrationlost` on `ServiceWorkerGlobalScope`
- Define `push` event using interface `PushEvent`
- Define the simple event `pushregistrationlost`

### DIFF
--- a/index.html
+++ b/index.html
@@ -130,34 +130,23 @@
         Terminology
       </h2>
       <p>
-        The following concepts and interfaces are defined in [[!HTML5]]:
+        <dfn><a href="http://www.w3.org/TR/html5/webappapis.html#queue-a-task">Queue a
+        task</a></dfn> and <dfn><a href=
+        "http://www.w3.org/TR/html5/webappapis.html#fire-a-simple-event">fire a simple
+        event</a></dfn> are defined in [[!HTML5]].
       </p>
-      <ul>
-        <li>
-          <dfn><a href="http://www.w3.org/TR/html5/webappapis.html#queue-a-task">queue a
-          task</a></dfn>
-        </li>
-        <li>
-          <dfn><a href="http://www.w3.org/TR/html5/webappapis.html#fire-a-simple-event">fire a
-          simple event</a></dfn>
-        </li>
-        <li>
-          <dfn><a href="http://www.w3.org/TR/html5/browsers.html#event-handlers">event
-          handler</a></dfn>
-        </li>
-        <li>
-          <dfn><a href="http://www.w3.org/TR/html5/browsers.html#event-handler-event-type">event
-          handler event type</a></dfn>
-        </li>
-      </ul>
       <p>
         <a href=
         'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise-objects'><dfn>Promise</dfn></a>
         is defined in [[!ECMASCRIPT]].
       </p>
       <p>
-        <dfn><a href="http://www.w3.org/TR/service-workers/#dnf-service-worker">Service
-        Worker</a></dfn> is defined in [[!SERVICE-WORKERS]].
+        <dfn>EventInit</dfn> is defined in [[!DOM]].
+      </p>
+      <p>
+        <dfn>Service Worker</dfn>, <dfn>ServiceWorkerRegistration</dfn>,
+        <dfn>ServiceWorkerGlobalScope</dfn>, and <dfn>ExtendableEvent</dfn> are defined in
+        [[!SERVICE-WORKERS]].
       </p>
       <p>
         The term <dfn>webapp</dfn> refers to a Web application, i.e. an application implemented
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -130,34 +130,23 @@
...
-          handler</a></dfn>
-        </li>
-        <li>
-          <dfn><a href="http://www.w3.org/TR/html5/browsers.html#event-handler-event-type">event
-          handler event type</a></dfn>
-        </li>
-      </ul>
       <p>
         <a href=
         'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise-objects'><dfn>Promise</dfn></a>
```
beverloo, line 140: nit: Filed #30 to track this link.

index.html
```diff
@@ -643,142 +632,86 @@
...
-      </p>
-      <p>
-        The <a>PushRegisterMessage</a> interface represents an event related to a <a>push
-        registration</a> that is no longer valid and thus needs to be created again by the
-        <a>webapp</a> by means of a new invocation of the <a><code>register</code></a> method.
-      </p>
-      <dl title="interface PushRegisterMessage" class="idl">
         <dt>
-          readonly attribute DOMString registrationId
+          attribute EventHandler onpushregistrationlost
```
beverloo, line 646: Is this name change agreed upon? I strongly prefer separating editorial changes from non-editorial ones.

index.html
```diff
@@ -643,142 +632,86 @@
...
+        </p>
+        <dl title=
+        "[Constructor(DOMString type, optional PushEventInit eventInitDict), Exposed=ServiceWorker] interface PushEvent : ExtendableEvent"
+        class="idl" data-merge="PushEventInit">
+          <dt>
+            readonly attribute DOMString? data
+          </dt>
+        </dl>
+        <p>
+          When getting the <code id="widl-PushMessage-data">data</code> attribute of a
```
beverloo, line 664: Tangentially related, I would be in favor of having a visible ISSUE somewhere in the spec referring to the mandatory encryption GitHub issue.

index.html
```diff
@@ -643,142 +632,86 @@
...
+          <code>PushEvent</code>, the <a>user agent</a> MUST return the message data received by
+          the <a>user agent</a> in the <a>push message</a>, or null if no data was received.
+        </p>
+        <dl title="dictionary PushEventInit : EventInit" class="idl">
+          <dt>
+            DOMString? data
+          </dt>
+        </dl>
+        <p>
+          Upon receiving a <a>push message</a> for a <a>webapp</a> from the <a>push server</a> the
```
beverloo: s/push server/push service/. Push Service is defined as the front-end an application server talks to, whereas this would in many cases be the independent backend.

index.html
```diff
@@ -643,142 +632,86 @@
...
+          <a>user agent</a> MUST run the following steps:
+        </p>
+        <ol>
+          <li>If the <a>Service Worker</a> associated with the <a>webapp</a> is not running, start
+          it.
```
beverloo, line 679: This needs a reference to the Service Worker specification. A concept of "starting a service worker" is too large to leave without a link.

index.html
```diff
@@ -643,142 +632,86 @@
...
+          </li>
+          <li>Let <var>scope</var> be the <code>ServiceWorkerGlobalScope</code> of the <a>Service
+          Worker</a> associated with the <a>webapp</a>.
+          </li>
+          <li>Let <var>event</var> be a new <a><code>PushEvent</code></a> whose <code>data</code>
+          attribute is the message data received by the <a>user agent</a> in the <a>push
```
beverloo, line 685: nit: Should we specify here that the message must have come from the Push Service, to be consistent with the text above?

index.html
```diff
@@ -643,142 +632,86 @@
...
+          </li>
+        </ol>
+      </section>
+      <section>
+        <h2>
+          The <code>pushregistrationlost</code> event
+        </h2>
+        <p>
+          Upon any event that makes a <a>push registration</a> no longer valid, e.g. <a>push
+          server</a> database failure, the <a>user agent</a> MUST run the following steps:
```
beverloo, line 700: The named example is rather unlikely to happen, isn't it? I filed #61 to keep track of what we should do when the user clears their cache.

index.html
beverloo, line 646: Seems like there is some consensus in #18. OK.

index.html
mvano, line 664: That seems like something for the security and privacy section, which this patch is really not about. I filed isse #62 to go over this.

index.html

index.html
mvano, line 685: That seems redundant.

index.html
mvano, line 700: The database loss example is unchanged from earlier versions of the spec. I'm not sure how a user agent would even be notified if the registration data was lost... how would the server know who to notify without the identifying data? I'd like to discuss that with the people that came up with this example.

index.html
mvano, line 679: I can't find a good description to reference in the Service Worker spec. I filed this issue for it: https://github.com/slightlyoff/ServiceWorker/issues/501

mvano, line 679: And for the Push API spec there is now issue #63.

PR #72, opened 2014-10-03
Clean up links that redirect and define the DOM bits.

mvano:
This makes the Link Checker happier so we can proceed with WD publication.


beverloo:
LGTM

PR #76, opened 2014-10-09
Nit: Clarify wording of "univocal" identifier

johnmellor:
None of the meanings on http://en.wiktionary.org/wiki/univocal made sense in this context; I think the original intention was that these identifiers would be canonical, i.e. you can compare them directly without having to normalize them first.

### DIFF
--- a/index.html
+++ b/index.html
@@ -633,7 +633,7 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
       <p>
         When getting the <code><dfn id=
         "widl-PushRegistration-registrationId">registrationId</dfn></code> attribute, the <a>user
-        agent</a> MUST return a univocal identifier of this <a>push registration</a> in the <a>push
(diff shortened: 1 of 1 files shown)


efullea:
univocal was meant to indicate that t he registrationIdunambiguously identify a push registration, that is, there is a 1-to-1 relationship between the push registration and the registrationId. So it matches with meaning number 5.

If univocal is not clear we may say: "MUST return an identifier that unambiguoslyh identifies this push registration"

But canonical was not the intended meaning

martinthomson:
I believe that the intended term is 'unequivocal'.  @johnmellor's proposal is clearer.

efullea:
No, the intended term was 'univocal' http://www.thefreedictionary.com/univocal'. For sure, I wrote it. It seems though that it is not well understood. 

'unequivocally' can also work if it is more clear.

PR #87, opened 2014-11-13
Add optional userVisibleOnly parameter to register & hasPermission

johnmellor:
See discussion at:
http://lists.w3.org/Archives/Public/public-webapps/2014OctDec/0214.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -414,9 +414,14 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
         "webapp">webapps</a> to establish access to <a title="push service">push services</a>. Note
         that just a single <a>push registration</a> is allowed per <a>webapp</a>.
       </p>
-      <dl title="interface PushRegistrationManager" class="idl">
+      <dl title="dictionary PushRegisterOptions" class="idl">
         <dt>
-          Promise&lt;PushRegistration&gt; register ()
+          boolean userVisibleOnly = false
+        </dt>
+      </dl>
+      <dl title="interface PushRegistrationManager" class="idl" data-merge="PushRegisterOptions">
+        <dt>
+          Promise&lt;PushRegistration&gt; register (optional PushRegisterOptions options)
         </dt>
         <dt>
           Promise&lt;PushRegistration&gt; unregister ()
@@ -425,13 +430,13 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
           Promise&lt;PushRegistration&gt; getRegistration ()
         </dt>
         <dt>
-          Promise&lt;PushPermissionStatus&gt; hasPermission ()
+          Promise&lt;PushPermissionStatus&gt; hasPermission (optional PushRegisterOptions options)
         </dt>
       </dl>
       <p>
         The <code><dfn id=
-        "widl-PushRegistrationManager-register-Promise-PushRegistration">register</dfn></code>
-        method when invoked MUST run the following steps:
+        "widl-PushRegistrationManager-register-Promise-PushRegistration--PushRegisterOptions-options">
+        register</dfn></code> method when invoked MUST run the following steps:
       </p>
       <ol>
         <li>Let <var>promise</var> be a new <a><code>Promise</code></a>.
@@ -444,7 +449,10 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
         </li>
         <li>Ask the user whether they allow the <a>webapp</a> to receive <a title=
         "push message">push messages</a>, unless a prearranged trust relationship applies or the
-        user has already granted or denied permission explicitly for this <a>webapp</a>.
+        user has already granted or denied permission explicitly for this <a>webapp</a>. This may
+        depend on the value of <a href=
+        "#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a> — see <a href=
+        "#widl-PushRegisterOptions-userVisibleOnly"></a>.
         </li>
         <li>If not granted, reject <var>promise</var> with a <code><a>DOMException</a></code> whose
         name is "<a href=
@@ -533,8 +541,8 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
       </ol>
       <p>
         The <code><dfn id=
-        "widl-PushRegistrationManager-hasPermission-Promise-PushPermissionStatus">hasPermission</dfn></code>
-        method when invoked MUST run the following steps:
+        "widl-PushRegistrationManager-hasPermission-Promise-PushPermissionStatus--PushRegisterOptions-options">
+        hasPermission</dfn></code> method when invoked MUST run the following steps:
       </p>
       <ol>
         <li>Let <var>promise</var> be a new <a><code>Promise</code></a>.
@@ -542,7 +550,9 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
         <li>Return <var>promise</var> and continue the following steps asynchronously.
         </li>
         <li>Retrieve the push permission status (<a><code>PushPermissionStatus</code></a>) of the
-        requesting <a>webapp</a>
+        requesting <a>webapp</a>. This may depend on the value of <a href=
+        "#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a> — see <a href=
+        "#widl-PushRegisterOptions-userVisibleOnly"></a>.
         </li>
         <li>If there is an error, reject <var>promise</var> with no arguments and terminate these
         steps.
@@ -559,6 +569,41 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
         If there is a need to ask for permission, it needs to be done by invoking the
         <a><code>register</code></a> method.
       </p>
+      <section id="widl-PushRegisterOptions-userVisibleOnly">
+        <h3>
+          Registering for userVisibleOnly Push
+        </h3>
+        <p>
+          If a <a>webapp</a> passes <code>true</code> for the <a href=
+          "#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a> parameter to
+          <a><code>register</code></a>, they are committing to show user-visible UI (such as a
+          notification) for every <a>push message</a> that the <a>webapp server</a> sends to the
+          <a>webapp</a>.
+        </p>
+        <p>
+          In exchange, the <a>user agent</a> MAY show different UI for the
+          <a><code>register</code></a> method (some <a title="user agent">user agents</a> may even
+          always reject with a "<a href=
+          "https://www.w3.org/Bugs/Public/show_bug.cgi?id=23033"><code>PermissionDeniedError</code></a>"
+          unless <a href="#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a> is
+          <code>true</code>). Other <a title="user agent">user agents</a> might simply ignore it.
+        </p>
+        <p>
+          However, <a title="user agent">user agents</a> MAY penalize <a title="webapp">webapps</a>
+          that pass <code>true</code> for <a href=
+          "#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a> then break the commitment
+          above (for example by throttling <a title="push message">push messages</a>, or showing a
+          generic notification to the user whenever the <a>webapp</a> fails to show user-visible UI
+          after receiving a <a>push message</a>).
+        </p>
+        <p>
+          If a <a>user agent</a> chooses to handle the <a><code>register</code></a> method
+          differently depending on the value of the <a href=
+          "#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a> flag, it MUST reflect
+          that appropriately when <a><code>hasPermission</code></a> is called with the same value
+          for <a href="#idl-def-PushRegisterOptions"><code>userVisibleOnly</code></a>.
+        </p>
+      </section>
       <section>
         <h3>
           Push Registration Persistence


martinthomson:
So we've had a few discussions about this, and while none of this is concrete, we're not convinced by the arguments presented for this particular UX design.  As such I'm willing to look into ways that we can facilitate Google's needs without creating an API surface that anyone would be required to commit to.

With the current form of the API, and the permissions interaction model it implies, I can appreciate most of the logic process that leads to your conclusions, but we still come to different conclusions.  I don't want to be in a position of dictating UX on Chrome, that's not healthy, but there are certain expectations that I think we need to establish.  The model that the current API imposes constraints on UX that could be incompatible with the nascent model that I have in mind based on the discussions I've had so far.

I'm going to talk to our UX designers and some privacy folks before I commit to something more concrete, but I'm currently inclined to propose a different consent model for this API than the one currently specified.  And just to assuage concerns about timeliness, I don't expect that to take long.

martinthomson:
I've shopped this idea, and variations on it, around a bit.  I've now talked to a bunch of people about what their expectations are and it's clearly not a straightforward decision.

I should say that I'm not responsible for making a final determination from a Firefox perspective; we're still very much in the process of working out what we want from this feature at the product level.

### First, the negative reactions

The reaction to a having multiple levels of consent grant for a single capability was generally negative.  I think that this was because it wasn't clear that it was possible to communicate the difference clearly.

Even in the conversations I had, it was often the case that the distinction between "push notifications" and "notifications" (i.e., toasts or equivalent) was confusing.  Language matters, so I've been using "run in the background" and "display notifications" as the names of capabilities we're talking about.  

I also found it hard to convey the full range of subtlety in the proposed UX to some people who are highly technically adept, so I worry that it might be a little hard to explain this to web developers and users.

### Consent and feedback

The overwhelming majority of people think that we should have some sort of consent, but that the consent for "running in the background" (hereafter, just "push") and "displaying notifications" need to be orthogonal.  How orthogonal varied though; it certainly wasn't clear cut.

The amount of concern over the ability to periodically trigger requests was - on the whole - low.  Though generally understood to be a concern, it wasn't considered to be qualitatively worse than the resource consumption concerns (i.e., access to CPU and storage).  As long as there is some elevation process from random drive-by origin, most people were OK providing push, though there were some exceptions.

One of the key points I took away was that feedback was hugely important, but the nature of that feedback might vary. Toasts, or visible notifications in that vein, are just one such option.  I had some fairly positive discussions about feedback in a more general sense, which I think is the right direction to be taking the discussion.

In particular, Anne van Kesteren had some good ideas and information.  There is no consent process for access to storage for sites.  I believe that current designs don't even enforce a quota (this might change). Instead, the idea there is to provide feedback in the form of a display, showing how much capacity each site uses.  Coupled with a notification when storage capacity crosses certain thresholds, users are left in control (they can clear state for a site), but they aren't burdened with consent choices.

A similar system of feedback for push would significantly reduce the load on the consent process.  Currently, the implication is that a bad choice exposes the user to indefinite tracking and expenditure of resources.  From that perspective, I can understand the desire to somehow limit this exposure.  A goal of clearly communicating what is going on and providing immediate controls is what I'm trying to head towards.  I don't want that to be only a long term aspirational goal though, because stop-gaps have a way of shaping the landscape, sometimes in negative ways.

### Options

Here are what I think the options are for the permission model.

#### 1. Graded permissions

Two levels of permission grant for push, as proposed.  The proposal elevates the two levels to become a first class part of the API.  What I'm hearing implied in this proposal is a desire to see if it is possible to gradually eliminate the real differences between the levels, but please correct me.

I'd like to see the expectations around each level - particularly this intermediate stage - be expanded considerably.  When we we discussing this, it seemed like the system of enforcement or penalties was still a little nebulous.  If this is going to hit the API, then it needs to be very clear what an application can expect from the browser when it opts for each level of permission.

#### 2. One permission

A single "run in the background" permission.

This option would be entirely orthogonal to the other permissions in the system.  If you want to generate visible notifications, then consent for the notifications API needs to be acquired separately.

#### 3. Some interaction with notification consent

There are a few variations on this one.  Basically, the idea is that if a site has acquired consent for this (related) feature, then it's OK.

Part of the rationale for these options is the close conceptual relationship between the push and notification features.  There is also a theory that users expect to receive notifications from sites when they aren't visiting them, and might be surprised to learn how infrequently this is actually true.

If you consider the two identified classes of permission we are talking about (push visible and push invisible), plus an existing consent for notifications, you can cover the space almost exhaustively:
- a. If you can generate notifications, then you can run in the background.  There is no separate permission.
- b. If you can generate notifications, you can run in the background, but if you don't generate a notification, the browser will generate a generic one for you.  Getting complete push access requires a separate permission, but you need to acquire notification consent separately.
- c. A push access permission implicitly grants consent to generate notifications.  The notifications permission only covers the case where an open window/tab generates a notification.
- d. b., but with push consent implying notification consent, like c. instead.

One person I talked to about UX was particularly in favour of ensuring that notifications are kept separate from background running, even if the notifications weren't being generated by an active window/tab.  Option b here sort of covers that.

#### (Extra) Options for limiting exposure

There are alternative approaches to reducing the cost to the user of a bad consent choice.  A consent grant doesn't have to be permanent.  If the use of a push notification service were limited in time or uses (or both), then any mistake is similarly non-permanent.

Any limited budget might be refreshed if the site is visited (i.e., the top-level browsing context hits the origin).

When the subscription expires, deliver a `pushregistrationlost` event to the SW that terminated the push subscription.  This event might need to be augmented to have some sort of indicator that the error is unrecoverable.

You can imagine getting arbitrarily inventive with the accounting of time, visible and invisible pushes.

### My preference

The general idea of having API surface as a protection measure still grates, though I can appreciate the concern.

I'm also concerned about creating a perverse set of incentives whereby applications generate user visible notifications when they aren't otherwise necessary or desired.  I don't think it is in the interest of users to have this sort of incentives in place.

I think that option 2 is simplest, but I quite like 3c.  Adding expiration of consent is very attractive; and it may not even need to remain a short-term feature.  Once we have good site status feedback UX in place, then the need for expiration might diminish, but it could turn out to be a good natural limiting function.

johnmellor:
Thanks for looking into this in depth. It sounds like at a high level, we are surprisingly well aligned :-)

> feedback [is] hugely important (...) the idea [for storage] is to provide feedback in the form of a display, showing how much capacity each site uses (...) A similar system of feedback for push would significantly reduce the load on the consent process. Currently, the implication is that a bad choice exposes the user to indefinite tracking and expenditure of resources.

We strongly agree. Chome is about to launch an improved [Permission/Resource Manager](https://docs.google.com/a/chromium.org/document/d/1oQwmj3AU4QYhTyGrYEGr6zaZhHUfx-wqUgEcQGbUU-U/edit), for example for storage you'll be able to see sites sorted by how much space they use.

For push/notifications, we'll show when the site last sent/showed a push/notification. But realistically, most users won't go looking at this UI. So the most effective feedback we have is for notifications, where each notification is an implicit reminder that the site is doing stuff in the background, and we'll make it easy to revoke permission from a notification.

> 3b. If you can generate notifications, you can run in the background, but if you don't generate a notification, the browser will generate a generic one for you. Getting complete push access requires a separate permission, but you need to acquire notification consent separately.

This is exactly the model we think makes sense.

> 3c. A push access permission implicitly grants consent to generate notifications. The notifications permission only covers the case where an open window/tab generates a notification.

I'm confused - why would push permission grant foreground notifications but not background notifications? Surely the point of using push is to show notifications in the background?

> A consent grant doesn't have to be permanent (...) Any limited budget might be refreshed if the site is visited (...) You can imagine getting arbitrarily inventive with the accounting of time, visible and invisible pushes.

Yes, our security UX team have been thinking along similar lines about automatically setting a dynamic quota for silent push messages based on user engagement with the site (with sufficiently good heuristics, it might even be possible to avoid showing confusing prompts at all for this use case, and let the UA handle the tedious task of managing resource consumption - we shall have to see).

Can I ask how you would you represent option 3b in the API? (this pull request was my attempt to do so, but it seems it has been confusing).

martinthomson:
> So the most effective feedback we have is for notifications, where each notification is an implicit reminder that the site is doing stuff in the background, and we'll make it easy to revoke permission from a notification.

We discussed this as well, and concluded that we couldn't rely on that alone (notification UX being outside our control on some platforms).  And limited real estate means that there may simply be too little space anyway.  Not saying that it isn't worth looking into though.

> > 3c. A push access permission implicitly grants consent to generate notifications.  The notifications permission only covers the case where an open window/tab generates a notification.
> > I'm confused - why would push permission grant foreground notifications but not background notifications?

It's the opposite: if you can do push, you can notify everywhere.  But the existing Notification permission is unchanged here.  The argument for this case is that we don't change the meaning of an existing consent grant, but merely add another one that provides a wider scope.  The important thing to note with this is that it doesn't distinguish between pushes that do and don't generate notifications.

And in case code would help:

``` js
Notifications.requestPermission().then(() => {
  // can show notification from regular content
  assert(Notifications.hasPermission());
  // can't run in background (i.e., web push)
  assert(!Background.hasPermission());
});
```

vs

``` js
Background.requestPermission().then(() => {
  assert(Notifications.hasPermission());
  assert(Background.hasPermission());
});
```

> Surely the point of using push is to show notifications in the background?

I don't really want to make too many assumptions about what people plan to do with the increased availability that push provides.  Most of the cases I come back to revolve around hosting and sovereignty issues, most of which can be addressed to a large extent with crypto if they don't require user involvement (and user involvement means notifications).

> Yes, our security UX team have been thinking along similar lines about automatically setting a dynamic quota for silent push messages based on user engagement with the site (with sufficiently good heuristics, it might even be possible to avoid showing confusing prompts at all for this use case, and let the UA handle the tedious task of managing resource consumption - we shall have to see).

That's where we are at already, mostly.  I think we may require some sort of indication from the user that a site is more trusted than any random origin.  I can speculate about how you might avoid a prompt altogether, but I'm not keen on the idea of allowing every site you visit - accidentally or otherwise -  to create background trackers.  If a scheme could be devised that can't be gamed by sites, let me know; we have plenty of other user consent prompts that we hate.

More likely, we'll start with a simple timeout (e.g., one or two weeks).  And then maybe move to a credit-based scheme that runs down with time and pushes.  I'm still disinclined to significantly discount pushes that become visible, but that probably needs a little more modelling.

> Can I ask how you would you represent option 3b in the API?

I'd simply have the two permissions orthogonal from an API perspective.  Rather than gate access to the SW on the "background" permission, I would also allow it when the "notification" permission was active (both can be queried and you can use the logical OR of the results to determine if everything is OK).  So that would result in no change to the API, just the descriptive text saying when the SW is permitted.

johnmellor:
**_tl;dr: In order for UAs to be free to implement any of 2, 3a, 3b, 3c or 3d etc, we need to have two different ways of requesting push permission. How about `PushManager.requestSilentPermission()` and `PushManager.requestUserVisiblePermission()`?**_

@martinthomson wrote:

> we couldn't rely on [notifications as a form of feedback] alone (notification UX being outside our control on some platforms). And limited real estate means that there may simply be too little space anyway. Not saying that it isn't worth looking into though.

It should be possible on Android 4.1+, Firefox OS, and several desktop platforms at least. On JellyBean+ we plan for notifications to always show the origin, and additionally show a Site Settings button when you expand a notification, and on Lollipop+ we'll additionally show a settings cog when you long-press on the notification. The button and cog will both deep-link to the content settings page for the origin that showed the notification, allowing you to revoke the permission.

> 3c. (...) It's the opposite: if you can do push, you can notify everywhere. But the existing Notification permission is unchanged here. The argument for this case is that we don't change the meaning of an existing consent grant, but merely add another one that provides a wider scope.

Gotcha, that makes more sense (especially with the code samples - thanks!).

> > Can I ask how you would you represent option 3b in the API?
> > I'd simply have the two permissions orthogonal from an API perspective. Rather than gate access to the SW on the "background" permission, I would also allow it when the "notification" permission was active 

But wouldn't spec'ing things that way prevent UAs from implementing 3c (since developers would incorrectly assume that having notification permission meant they would be able to use push notifications)?

We can debate for days which model is the best tradeoff, but my goal is to get to a place where the specs allow UAs to implement any of 2, 3a, 3b, 3c or 3d etc (and evolve this over time) yet have the same JS work in all browsers (without unnecessary prompts).

Suppose we had the following strawman API:

``` js
partial enum PermissionName {
    "notifications",
    "user_visible_push_messaging",
    "silent_push_messaging",
    "background_sync"
}
[Exposed=Window]
partial interface Permissions {
    // Returns a sequence of Promises, since some may be resolved earlier than
    // others if multiple prompts are shown.
    static sequence<Promise<PermissionState>> request(PermissionName... name);
}
```

Then ideally each of the following JS snippets would "do the right thing" in all UAs:

``` js
// A. Notifications shown from a foreground window/worker only (though the
// notifications themselves may be persistent).
Permissions.request("notifications").then(...);

// B. Notification-only Push (some browsers will force these to show a
// notification for almost every push).
Promise.all(Permissions.request("notifications",
                                "user_visible_push_messaging")).then(...);

// C. A mix of silent push and push notifications (browsers will never force
// these to show a notification).
Promise.all(Permissions.request("notifications",
                                "user_visible_push_messaging",
                                "silent_push_messaging")).then(...);

// D. Silent push messaging.
Permissions.request("silent_push_messaging").then(...);

// E. Background sync.
Permissions.request("background_sync").then(...);

// F. Silent push and background sync.
Promise.all(Permissions.request("silent_push_messaging",
                                "background_sync")).then(...);

// G. Gimme everything!
Promise.all(Permissions.request("notifications",
                                "user_visible_push_messaging",
                                "silent_push_messaging",
                                "background_sync")).then(...);
```

Specifically if Chrome chooses to implement 3b:
- A, B would show a push notifications prompt
- D, E, F would show a background prompt
- C, G would show both a notifications and a background prompt (either combined, or one after the other)

And if Firefox chooses to implement 2:
- A would show a notifications prompt
- D, E, F would show a background prompt
- B, C, G would show both a notifications and a background prompt (either combined, or one after the other)

Or if Firefox chooses to implement 3c:
- A would show a notifications prompt
- B, C, D, E, F, G would show a background prompt

Another UA might choose to grant background_sync for free to any website that asks for it, yet prompt for silent_push_messaging (on grounds that the UA can throttle background_sync at will, but can't throttle silent_push_messaging without dropping possibly important messages).

However such a strawman extension to the [Permissions API](https://w3c.github.io/permissions/), would be a rather large dependency; in the meantime, the same effect could be achieved with existing APIs using independent\* method calls like `Notification.requestPermission()` to request each of notifications, user_visible_push_messaging, silent_push_messaging and background_sync. It's already possible for browsers to detect when several permissions are synchronously requested without returning to the message loop (permission requests don't block each other of course, since they're all async), and in such cases they can coalesce the prompts however they choose.

With the currently spec'ed APIs, it's thus already possible to implement 2, 3a, or 3c. Unfortunately UAs can't yet implement 3b or 3d, since there's no distinction between user_visible_push_messaging and silent_push_messaging. Hence this pull request seeks to introduce such a distinction, allowing all five policies to be implemented.

It seems you're not keen on the current patch's approach of treating user-visible-push and both-silent-and-user-visible-push as two different levels of the same permission. Would it be better to explicitly split them out as follows?:

``` js
interface PushManager {
    Promise<PushRegistration> register ();
    Promise<PushRegistration> getRegistration ();
    Promise<PermissionStatus> requestSilentPermission ();
    Promise<PermissionStatus> requestUserVisiblePermission ();
    Promise<PermissionStatus> hasSilentPermission ();
    Promise<PermissionStatus> hasUserVisiblePermission ();
};
```

Then for example developers would implement G. Gimme everything as:

``` js
Promise.all([
    Notification.requestPermission(), // Technically this doesn't yet return a Promise
    pushManager.requestSilentPermission(),
    pushManager.requestUserVisiblePermission(),
    syncManager.register(id, options) // We could add a requestPermission method to the spec instead
]).then(...);
```

*: (To ensure that authors write JS that will work in all UAs, rather than hardcoding the current UX choices of the browser(s) they test in, UAs should probably return "prompt" when a permission has not yet been asked for, even if previously granted permissions mean the prompt will be automatically granted without bothering the user. We should probably spec that.)

martinthomson:
Unfortunately, I think that an A or B or C or D option creates more problems.

I don't believe that the permissions API, as specified, is a good idea.  Your proposal seems predicated on the notion that the consent model implied by the permissions API is the right way to address the general class of problems that we as browser developers don't yet know how to solve.  I'll acknowledge that there are problems with the current scheme for acquiring consent, but that those are not the ones that this model aims to address.  The proposed spec in fact avoids addressing any such concerns, it merely _is_.  (This is my first encounter with that proposal, so maybe there is some context I'm missing.)

I'd expect a longer discussion about what it means to obtain consent as a preface to any work on improving permissions.  For instance, see [this post by Robert O'Callahan](http://robert.ocallahan.org/2011/06/permissions-for-web-applications_30.html), which sums up one view on this fairly tidily.  The main advantage of the proposed permissions API seems to care more about streamlining the application and browser developer experience than it does about engaging on serving the needs of users.

Secondly, the creation of a language for the expression of more granular preferences doesn't help improve the consistency of a user experience.  The opposite is more likely true: different browsers can build consent interaction models that are not just superficially different, but fundamentally different.

We've seen that even superficial differences in terms of the placement of dialogs has caused significant pain for applications using [getUserMedia](https://w3c.github.io/mediacapture-main/) (which is notably absent from permissions).  There, what amounts to user agent sniffing is commonplace.  The only reason it's not a full-blown inspection of `navigator.userAgent` is that applications are able to use the prefix on the API to detect which browser it is.

Finally, I would rather have a model closer to the "ask forgiveness, not permission" for this particular grant.  The distinction between stealthy and overt is only that way because we haven't got good accountability for the former in place.  And I'm slowly coming around to the view that - once some form of consent is granted - the distinction is basically moot.  

If you would like that to be more concrete, I'm tending more toward 3a; as a simplification of 3c, it is quite attractive.  And that probably wasn't what you were looking for.

johnmellor:
I think my mention of possible extensions to the Permissions API has been a red herring :-)

@martinthomson wrote:

> see [this post by Robert O'Callahan](http://robert.ocallahan.org/2011/06/permissions-for-web-applications_30.html)

Yes, we've all read that post, and agree pretty much word for word. Like Robert, we dislike asking upfront, and dislike large bundles of permissions.

However notifications and push (and in future background sync) are unique in two ways:
1. They can be more user-comprehensible when explained together.
2. Crucially, they operate in the background, so we can't force sites to ask for permission in context (by making the permission request implicit in the API), instead we have to allow prompting upfront. Notifications has embraced this, and has an explicit requestPermission method (with [a warning](https://notifications.spec.whatwg.org#static-members) that this is unusual). We could do the same for Push (as suggested in my last comment), though it seems slightly nicer to keep the permission request as an implicit part of registration, as is done by the original pull request.

> I'm tending more toward 3a; as a simplification of 3c, it is quite attractive.

Ok, but since Chrome UX prefers 3b, can we find some way for the spec to allow UAs to implement either?

martinthomson:
> allow UAs to implement either?

I think that you missed my second point in the noise.  It's damaging - to all of us - to have divergence on this.

Is there any reason you can't simply tune your revocation heuristic to severely penalize non-notifiers in the short term and maybe automatically reject the creation of the SW if the site can't generate notifications?

johnmellor:
@martinthomson wrote:

> I'm tending more toward 3a (...)
> a. If you can generate notifications, then you can run in the background. There is no separate permission.

We've continued to discuss this internally, and have three largish concerns with 3a:
1. Our user studies showed that users who grant notifications do not expect it to allow apps to run in the background, and that is a thing they are concerned about (for battery/creepiness/privacy reasons).
2. Users who would like e.g. news sites to sync in the background may deny notification permission if the site has a reputation for spammy advertising.
3. If users don't realise they're granting silent push, their location can be geoip tracked without their awareness (whereas if notifications are shown, the user at least knows the app is doing something, so it's comparable to open tabs tracking your IP).

> It's damaging - to all of us - to have divergence on this.

Convergence would be nice, but the Chrome Security UX team is convinced that 3b is most understandable for users; if we can't agree on that, it seems we'll have to start off with different UI and hopefully later converge once data proves which is more effective.

We'd still like the spec to allow for UAs to choose 3b (without excluding the alternatives). If we can't agree on that either, our current short-term strawman is to ask developers to put "gcm_user_visible_only":true in their manifest if they opt in to notification-only push, otherwise we will treat calls to PushManager.subscribe() as requests for silent push messaging (but a developer that has separately called Notification.requestPermission() would of course be allowed to show notifications as well).

slightlyoff:
I'm deeply concerned by the overall tone of this thread.

It is inappropriate to be specing UI details in non-layout APIs. In particular, the use of Promises in APIs like this is _explicitly_ to enable UA's to make potentially divergent runtime decisions about what sorts of UI (if any) to provide about security decisions.

It seems that the assumption in some of the discussion here is that agreement about UI is either helpful or desirable. It is not.

UAs must be able to retain autonomy about how/when to show security UI, what to put in it, and what sorts of options to present to users. Arguments against API because of optional (and likely-to-change) UI treatments or layered semantic restrictions are non-arguments unless they coincide with some other semantic issue.

Mozilla, for instance, has frequently requested descriptive strings in APIs even though Chrome will not use them. These are fine things to add to APIs assuming no demands are made on eventual UI (which, given form-factor changes alone, cannot hold). Reciprocity must be provided.

`hasPermission()` and extensions to it to deal with states Mozilla may not implement (e.g. `userVisibleOnly`) might have downsides, but they're strictly better than not providing the ability to test out these systems and get feedback. They should be added post-haste to help developers reason about the state of the system in their apps and predict and optimise the points at which requests might be user-visible. 

We seek a flexible model that allows us to iterate over time. Arguments about dialog placement for `getUserMedia()` are at best specious: we do not seek a consistent user experience across browsers. We want a consistent, safe experience on each browser and want a diversity of views tested so that, over time, we will evolve toward a better-performing UX.

If your team has data about why you think your UX decisions (or ours) are good enough to justify baking them in at this point, would love to see it. Until then, flexibility must be preserved. Attempts to get agreement about what sort of consent model to coalesce on without data are destined to failure.

martinthomson:
@johnmellor I think that we are ok to close this, right?

johnmellor:
Yup, thanks!

Issue #83, opened 2014-10-16
Strings or binary or both


martinthomson:
Push messages are severely space constrained (for good reason).  This is worse if we adopt mandatory encryption (and we should).

This creates an incentive to find ways of packing data as densely as possible.  That means binary payloads are quite desirable.  Do we want to enable binary data delivery?

Obviously this creates some complexity for applications.  Strings are still the most accessible form for messages.  I see several options:
1. binary data only
2. deliver both, with a string value only populated if it can be decoded successfully
3. deliver either and deliver a signal in the protocol, just like in thewebsocketprotocol or WebRTC

My initial leaning is toward 2.  I despise 3, though recognize that it may be necessary.

costinm:
HTTP (and every other protocol) handles this with Content-Type or similar meta-information. 
I'm not sure having the UA guess if the decrypted payload is a string is most efficient / robust - 
I would suggest meta-data, or 1 (binary data) plus some helper to decode it to string. 

Most apps know what their servers are supposed to be sending.

rektide:
Fetch API has a Body mixin that has an example, where the data is reified once and only once via a Promise (calls to different accessors fail). This might be a potential model for number 3, either.
https://fetch.spec.whatwg.org/#body-mixin

johnmellor:
Binary payload + Content-Type header + Fetch API's Body mixin for easy decoding to string / JSON / etc as appropriate sounds great.

(having the payload be a DOMString was a kludge caused by there being nothing as simple as the Body mixin when we designed it; but no-one was really keen on requiring Base64 or similar for binary data)

How would we actually expose the Content-Type header? As a Fetch API [Headers](https://fetch.spec.whatwg.org/#headers) object? I guess that'd mean we have to standardize the meaning of various headers in a push context.

martinthomson:
That might work.  If we can just lift the work that Fetch does, that's even better.

This is, however, somewhat incompatible with the proposed security enhancements (see #55) in that the metadata @costinm is describing will be on the outer envelope and cannot describe the actual message contents, post decryption.

costinm:
The outer envelope is generated by the Push server and/or UA - it is mostly
independent of the encrypted payload.
The encryption is intended ( among other things ) to prevent the push
server to access or modify the content of the message.
The metadata provide information from the push server, related to the
delivery and origin of the message.

martinthomson:
[...]

PR #88, opened 2014-11-17
Make PushEvent's data payload binary instead of just a string

johnmellor:
This exposes binary data as a Body mixin from the Fetch spec,
as discussed on https://github.com/w3c/push-api/issues/83

### DIFF
--- a/index.html
+++ b/index.html
@@ -159,19 +159,19 @@
       </p>
       <p>
         <code><a href=
-        'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise-objects'><dfn>Promise</dfn></a></code>
-        is defined in [[!ECMASCRIPT]].
+        'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise-objects'><dfn>Promise</dfn></a></code>,
+        and <code><a href=
+        "https://people.mozilla.org/~jorendorff/es6-draft.html#sec-json.parse"><dfn>JSON.parse</dfn></a></code>
+        are defined in [[!ECMASCRIPT]].
       </p>
       <p>
         <code><a href="http://www.w3.org/TR/dom/#eventinit"><dfn>EventInit</dfn></a></code>,
         <code><a href="http://www.w3.org/TR/dom/#domexception"><dfn>DOMException</dfn></a></code>,
         <code><a href="http://www.w3.org/TR/dom/#aborterror"><dfn>AbortError</dfn></a></code>,
         <code><a href=
-        "http://www.w3.org/TR/dom/#notfounderror"><dfn>NotFoundError</dfn></a></code>,
-        <code><a href=
-        "http://www.w3.org/TR/dom/#securityerror"><dfn>SecurityError</dfn></a></code>, and <a href=
-        "http://www.w3.org/TR/dom/#concept-event-constructor"><dfn>steps for constructing
-        events</dfn></a> are defined in [[!DOM]].
+        "http://www.w3.org/TR/dom/#notfounderror"><dfn>NotFoundError</dfn></a></code>, and
+        <code><a href="http://www.w3.org/TR/dom/#securityerror"><dfn>SecurityError</dfn></a></code>
+        are defined in [[!DOM]].
       </p>
       <p>
         <dfn>Service Worker</dfn>, <dfn>ServiceWorkerRegistration</dfn>,
@@ -179,17 +179,17 @@
         [[!SERVICE-WORKERS]].
       </p>
       <p>
-        <code><a href="https://fetch.spec.whatwg.org/#body"><dfn>Body</dfn></a></code>,
-        <code><a href="https://fetch.spec.whatwg.org/#bodyinit"><dfn>BodyInit</dfn></a></code>,
-        <a href="https://fetch.spec.whatwg.org/#concept-bodyinit-extract"><dfn>extracting</dfn></a>
-        a BodyInit, and the Body mixin's <a href=
-        "https://fetch.spec.whatwg.org/#concept-body-body"><dfn>associated body</dfn></a> are
-        defined in [[!FETCH]].
+        The algorithm <a href="http://www.w3.org/TR/encoding/#utf-8-decode"><dfn>utf-8
+        decode</dfn></a> is defined in [[!ENCODING]].
+      </p>
+      <p>
+        <code><a href="http://www.w3.org/TR/FileAPI/#blob"><dfn>Blob</dfn></a></code> is defined in
+        [[!FILEAPI]].
       </p>
       <p>
         <code><a href=
-        "https://heycam.github.io/webidl/#common-BufferSource"><dfn>BufferSource</dfn></a></code>
-        is defined in [[!WEBIDL]].
+        "http://heycam.github.io/webidl/#idl-ArrayBuffer"><dfn>ArrayBuffer</dfn></a></code> is
+        defined in [[!WEBIDL]].
       </p>
       <p>
         The term <dfn>webapp</dfn> refers to a Web application, i.e. an application implemented
@@ -657,6 +657,49 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
     </section>
     <section>
       <h2>
+        <code>PushMessageData</code> interface
+      </h2>
+      <dl title="interface PushMessageData" class="idl">
+        <dt>
+          ArrayBuffer arrayBuffer ()
+        </dt>
+        <dt>
+          Blob blob ()
+        </dt>
+        <dt>
+          JSON json ()
+        </dt>
+        <dt>
+          USVString text ()
+        </dt>
+      </dl>
+      <p>
+        Objects implementing <a><code>PushMessageData</code></a> gain an associated
+        <dfn>bytes</dfn> (a byte sequence, initially empty).
+      </p>
+      <p>
+        The <code><dfn id="widl-PushMessageData-arrayBuffer-ArrayBuffer">arrayBuffer()</dfn></code>
+        method, when invoked, MUST return an <a><code>ArrayBuffer</code></a> whose contents are
+        <var>bytes</var>.
+      </p>
+      <p>
+        The <code><dfn id="widl-PushMessageData-blob-Blob">blob()</dfn></code> method, when
+        invoked, MUST return a <a><code>Blob</code></a> whose contents are <var>bytes</var> and
+        <var>type</var> is not provided.
+      </p>
+      <p>
+        The <code><dfn id="widl-PushMessageData-json-JSON">json()</dfn></code> method, when
+        invoked, MUST return the result of invoking the initial value of
+        <a><code>JSON.parse</code></a> with the result of running <a>utf-8 decode</a> on
+        <var>bytes</var> as argument. This may throw an exception.
+      </p>
+      <p>
+        The <code><dfn id="widl-PushMessageData-text-USVString">text()</dfn></code> method, when
+        invoked, MUST return the result of running <a>utf-8 decode</a> on <var>bytes</var>.
+      </p>
+    </section>
+    <section>
+      <h2>
         Events
       </h2>
       <p>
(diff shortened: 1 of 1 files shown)


johnmellor:
@annevk, this pull request tries to reuse the [Body mixin](https://fetch.spec.whatwg.org/#body-mixin) defined by the Fetch spec for the [binary payload](https://cdn.rawgit.com/johnmellor/push-api/2cf8bb669e14bcd8099a8439142caf225cae0589/index.html#the-push-event) delivered by the Push API, since it's a clean API for consuming binary data in the format of the developer's choice. Does this seem reasonable? If so, two additional questions:
- It ended up being quite wordy; is there a clearer / more concise way for me to say unlike in normal event construction, the Body attribute is initialised from the BodyInit member?
- Is it ok that I'm directly exposing an attribute of type Body? It seemed simpler than defining an empty interface that simply implements Body. Alternatively I could make PushEvent directly implement the Body mixin (like Response does), but that might increase the risk of future namespace conflicts?

johnmellor:
HTML preview with this pull request applied: https://cdn.rawgit.com/johnmellor/push-api/binarydata/index.html

annevk:
@johnmellor does a push message really represent a stream? I thought it was just a single blob of data?

johnmellor:
@annevk interesting! A push is indeed a single binary blob (usually no more than 4 KB). Are you suggesting that e.g. Blob or ArrayBuffer would be more suitable?

It's important for accessing it as a utf-8 string to be easy, since that's the most common case. These days I guess the simplest way to decode an ArrayBuffer is: `new TextDecoder().decode(arrayBuffer)`?

It might still be easier for developers to have a Body-like interface with methods like `text()`, `arrayBuffer()`, `json()` etc which apply various decodings to the data. Though ideally, I suppose, such methods would already be present on ArrayBuffer itself.

annevk:
If this is not going to stream I recommend designing a simple `PushResponse` class that exposes `text()`, `json()`, etc. defined the same way as on `Response`/`Body`. Except these methods would be synchronous (but you can still only use one of them).

johnmellor:
@annevk wrote:

> but you can still only use one of them

Is this just for consistency with Body? If this is no longer a stream, it seems it would be more natural to allow it to be consumed multiple times (and garbage collected once unreferenced).

I've updated this pull request to do that, with a new [PushMessageData interface](https://rawgit.com/johnmellor/push-api/binarydata/index.html#pushmessagedata-interface). Does that look better?

Updated HTML preview: https://rawgit.com/johnmellor/push-api/binarydata/index.html

Thanks for the advice!

annevk:
The idea was allowing the UA to free the memory just as with `Body`, but since these are 4KiB in size that's probably silly.

annevk:
"Objects implementing PushMessageData gain an associated bytes (a byte sequence, initially empty)." -> "PushMessageData objects have an associated bytes" (There are no other objects implementing this.)

Also, in PushEventInit you probably want to mark the member as required. Otherwise it would be undefined what `new PushEvent("x").data` is.

johnmellor:
@annevk done, thanks.

index.html
```diff
@@ -675,26 +733,15 @@
           The <a>PushEvent</a> interface represents a received <a>push message</a>.
         </p>
         <dl title=
-        "[Constructor(DOMString type, optional PushEventInit eventInitDict), Exposed=ServiceWorker] interface PushEvent : ExtendableEvent"
+        "[Constructor(DOMString type, PushEventInit eventInitDict), Exposed=ServiceWorker] interface PushEvent : ExtendableEvent"
         class="idl" data-merge="PushEventInit">
           <dt>
-            readonly attribute DOMString? data
+            readonly attribute PushMessageData data
```
beverloo, line 775: If PushMessageData is required, how does the developer construct it? You should define the expected constructor arguments.

johnmellor, line 775: I can allow them to be constructed from a [BufferSource](http://heycam.github.io/webidl/#common-BufferSource). But I'm not sure whether that's useful, since IIRC you can't yet dispatch custom events to Service Workers?

I guess there's no harm, other than it might slightly restrict our ability to change the constructor in future (e.g. to add a type parameter).

index.html
```diff
@@ -675,26 +733,15 @@
...
-        </p>
-        <p>
-          When getting the <code id="widl-PushEvent-data">data</code> attribute of a
-          <code>PushEvent</code>, the <a>user agent</a> MUST return the message data received by
-          the <a>user agent</a> in the <a>push message</a>, or null if no data was received.
-        </p>
         <dl title="dictionary PushEventInit : EventInit" class="idl">
           <dt>
-            DOMString? data
+            required PushMessageData data
```
mvano: Why is PushMessageData not nullable? It's permitted to omit a payload when sending a push message. How would one check that the data is empty? There's no isEmpty method or something like that.

martinthomson: I imagine that `data.text().length === 0` will suffice.

annevk: Yeah, response.text(v => v === "") is also how it works in Fetch (cannot be synchronous in Fetch). In Fetch we might expose the raw body as a stream at some point and at that point we will probably allow distinguishing between empty and missing (just like HTTP).

For a new protocol it might be best to define empty and missing as having identical semantics so APIs can be simplified somewhat.

index.html
```diff
@@ -642,6 +664,100 @@
     </section>
     <section>
       <h2>
+        <code>PushMessageData</code> interface
+      </h2>
+      <div class="idl" title="typedef (Blob or BufferSource or USVString) PushMessageDataInit">
+      </div>
+      <dl title=
+      "[Constructor(PushMessageDataInit init), Exposed=ServiceWorker] interface PushMessageData"
```
beverloo: There's no reason to have a separate init typedef here (this mostly is used for events) -- the union could be placed in-line.

However, I'd argue that it should be fine to only support USVString as the constructor's parameter for now to avoid complicating this, until we learn that there's actually a need for multiple types. The PushMessageData structure is mostly intended for use by the engine (though it's not unreasonable to expect developers to write tests using them), and I wouldn't want to see people start using it as a generic converter between {Blob, ArrayBuffer, USVString, ...}.

beverloo:
@efullea @nikhilm Do you have any opinions about this change? This was one of the subjects we discussed in Madrid.

index.html
annevk: Perhaps we should have a generic object like that and use it here? Seems rather neat.

index.html
```diff
@@ -642,6 +666,93 @@
     </section>
     <section>
       <h2>
+        <code>PushMessageData</code> interface
+      </h2>
+      <dl title="[NoInterfaceObject] interface PushMessageData" class="idl">
```
beverloo, line 670: This shouldn't be a NoInterfaceObject. Per WebIDL:

"The [NoInterfaceObject] extended attribute should not be used on interfaces that are not solely used as supplemental interfaces, unless there are clear Web compatibility reasons for doing so."

index.html
```diff
@@ -642,6 +666,93 @@
...
+        The <code><dfn id="widl-PushMessageData-json-Any">json()</dfn></code> method, when invoked,
+        MUST return the result of invoking the initial value of <a><code>JSON.parse</code></a> with
+        the result of running <a>utf-8 decode</a> on <var>bytes</var> as argument. This may throw
+        an exception.
+      </p>
+      <p>
+        The <code><dfn id="widl-PushMessageData-text-USVString">text()</dfn></code> method, when
+        invoked, MUST return the result of running <a>utf-8 decode</a> on <var>bytes</var>.
+      </p>
+      <div class="idl" title="typedef (Blob or BufferSource or USVString) PushMessageDataInit">
```
beverloo: This is really complicated, and would still allow people to use this as a generic converter, just a little bit less obvious.

var event = new PushEvent('PushEvent', 'foobarbaz');
event.data.blob(); // ...

Given that this is a very minor detail, and with very(!) low expected usage, can we simplify this take a string, adding other types if there's a demonstrable need? For now, I'd be concerned about the added complexity.

johnmellor: Sure, I guess we can re-add the other PushMessageDataInit types later. Done.

index.html
johnmellor: @annevk where might we define such a generic object? WebIDL? Could we just make ArrayBuffer more powerful? I've [made PushMessageData a NoInterfaceObject](https://github.com/johnmellor/push-api/commit/b05231cce0f3415992d2dfeec290a6d174d0ef29) for now so that it could be swapped out for such a generic object in future. Does that sound reasonable? @beverloo's [not convinced](https://github.com/w3c/push-api/pull/88#discussion_r20810189) that it's worth it.

index.html
johnmellor, line 670: See discussion: https://github.com/w3c/push-api/pull/88#discussion_r20811259

index.html
annevk, line 775: Service workers must have `EventTarget` and therefore will have `dispatchEvent()`.

index.html
```diff
@@ -177,6 +181,26 @@
...
+        The algorithms <a href="http://www.w3.org/TR/encoding/#utf-8-encode"><dfn>utf-8
+        encode</dfn></a>, and <a href="http://www.w3.org/TR/encoding/#utf-8-decode"><dfn>utf-8
+        decode</dfn></a> are defined in [[!ENCODING]].
+      </p>
+      <p>
+        <code><a href="http://www.w3.org/TR/FileAPI/#blob"><dfn>Blob</dfn></a></code> is defined in
+        [[!FILEAPI]].
+      </p>
+      <p>
+        <code><a href="https://heycam.github.io/webidl/#idl-any"><dfn>Any</dfn></a></code>,
```
mvano: http://www.w3.org/TR/WebIDL/#idl-any

index.html
```diff
@@ -642,6 +666,86 @@
     </section>
     <section>
       <h2>
+        <code>PushMessageData</code> interface
```
mvano: Why isn't this <a>PushMessageData</a> like in the rest of this document?

index.html
```diff
@@ -642,6 +666,86 @@
...
+      <p>
+        The <code><dfn id="widl-PushMessageData-blob-Blob">blob()</dfn></code> method, when
+        invoked, MUST return a <a><code>Blob</code></a> whose contents are <var>bytes</var> and
+        <var>type</var> is not provided.
+      </p>
+      <p>
+        The <code><dfn id="widl-PushMessageData-json-Any">json()</dfn></code> method, when invoked,
+        MUST return the result of invoking the initial value of <a><code>JSON.parse</code></a> with
+        the result of running <a>utf-8 decode</a> on <var>bytes</var> as argument. This may throw
+        an exception.
```
mvano: What kind of exception? Where is this defined?

index.html
```diff
@@ -642,6 +666,86 @@
...
+        <a title="extract a byte sequence">algorithm</a> below would be extended, rather like the
+        Fetch standard's <a href="https://fetch.spec.whatwg.org/#concept-bodyinit-extract">extract
+        a byte stream</a> algorithm.
+      </p>
+      <p>
+        To <dfn>extract a byte sequence</dfn> from <var>object</var>, run these steps:
+      </p>
+      <ol>
+        <li>
+          <p>
```
mvano: Why are there paragraphs in these list items?

index.html
```diff
@@ -708,15 +801,28 @@
...
           </li>
           <li>
             <a>Queue a task</a> to <a title="fire a simple event">fire <var>event</var> as a simple
             event</a> named <code>push</code> at <var>scope</var>.
           </li>
         </ol>
+        <p>
+          When a constructor of the <a><code>PushEvent</code></a> interface, or of an interface
+          that inherits from the <a><code>PushEvent</code></a> interface, is invoked, the usual
+          <a>steps for constructing events</a> MUST be modified as follows: if the
```
mvano: Why if? The data attribute is now required. What happens if it is not set?

index.html
mvano, line 775: I think PushEventInit can remain optional.

index.html
johnmellor: Done (well, `<a>PushMessageData</a>`). Ditto everywhere else I'd used `<a><code>` for same-document IDL references).

index.html
johnmellor: Done.

index.html
johnmellor: I rephrased as `Re-throw any exceptions thrown by <a><code>JSON.parse</code></a>`.

index.html
johnmellor: Removed.

index.html
johnmellor: Hmm. Ok, I've made eventInitDict and it's data member optional again. bytes will just be an empty byte sequence if they aren't set.

index.html
johnmellor, line 775: I made PushEventInit optional again. It does indeed feel more natural.

mvano:
Looks good to me.

PR #89, opened 2014-11-24
Add and Encryption Key array to the PushRegistration interface

miguelgarciaarribas:


index.html
```diff
@@ -639,6 +645,17 @@
...
+      <p>
+        When getting the <code><dfn id=
+        "widl-PushRegistration-encryptionKeys">encryptionKeys</dfn></code> attribute, the <a>user
+        agent</a> MAY return one or more encryption keys.
+      </p>
+      <p class="note">
+        In order to encrypt a message the web developer would use
+        <a href="http://www.w3.org/TR/WebCryptoAPI/#dfn-KeyAlgorithm"> Key.algorithm </a> to detect what encryption algorithm
+        (and any necessary parameters) each key is for, then export the key (to upload to their server) in the KeyFormat
+        of their choice using crypto.subtle.exportKey(KeyFormat, Key).
```
johnmellor, line 657: Nit: please add a deeplink to http://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-exportKey

(and remove the spaces around Key.algorithm above)

johnmellor:
HTML preview: https://rawgit.com/miguelgarciaarribas/push-api/encryption_key/index.html

ShijunS:
The following is what I learned from the IETF session.
- The UA is expected to manage the private key of the key pair and handle the decryption task rather than leaving that to the webapp or its SW.  
- Otherwise, the webapp or its SW need to access the IndexedDB, and which in turn potentially leads to more operations.
- What the webapp does is to pass the public key to the webapp server.

If my understanding above is correct, we have another question.  WebCrypto might not provide the functionality for this.  The key pair in this case will be from and managed by the UA or the underlying system, which is not available from WebCrypto.  It seems to me the intention here is to define a new encryption behavior.  Of course, WebCrypto can always be a good reference in terms of algorithms and logics.

martinthomson:
This is a good start, but it isn't quite complete.  The spec also needs to define what key usage the keys will have, and what types they will include.

The key usage is simple.  We are minting a new key usage for WebRTC that has similar properties.  See w3c/webrtc-pc#28 for details.  I think that the right thing here is to define a "webpush" key usage with similar properties to the "webrtc" usage: i.e., you can look at the public key, but it is otherwise unusable by the web application.

The type or (algorithm, to be more precise) is harder.  Restricting this to any key usage that would ordinarily be good for the "decrypt" operation seems like the obvious choice, but in practice, the "deriveBits" operation is a better choice.  In WebCrypto terms, that means enabling ECDH and (Finite-Field) Diffie-Hellman.  Picking the one true algorithm that everyone has to implement is hard, because not only does it include that choice, but it includes groups or curves, plus symmetric protection as well.

index.html
martinthomson, line 657: While exportKey is needed, we need to avoid it on the private portion of this pair.  The private portion MUST NOT be `extractable`.

miguelgarciaarribas:
Key Usage sounds good, I'll take a stab at defining that.
I would prefer not to declare the exact algorithm here but point to whatever recommendation we decide at the IETF protocol, since as you say above at the JavaScript side of things the developer can only look at the key, not really do anything with it at the web level.

Does that sound reasonable?

martinthomson:
Yes, leave that to the IETF doc.

ShijunS:
Another question to consider in the W3C spec - given the key pair generation and the decryption operations are handled by the UA, should the UA be allowed to use the same key pair across all websites?  The public key is in the clear anyway, so I believe it won't make a lot of difference to give each site a different public key (and then manage all the key pairs in the UA).

martinthomson:
Every origin absolutely requires a different key pair, otherwise we generate a tracking vector.  My recommendation would be to generate one for each new push subscription, but that depends largely on choosing something relatively cheap to create, like ECDH.

johnmellor:
@martinthomson wrote:

> My recommendation would be to generate one for each new push subscription, but that depends largely on choosing something relatively cheap to create, like ECDH.

We should probably leave it up to the UA to decide whether to use a key per origin, or per Service Worker, or per push registration (or some other heuristic). The spec can then simply tell authors to make no presumptions about key re-use (i.e. authors should expect to have to store a separate key for every push registration, even if in some cases these may be the same).

martinthomson:
@johnmellor absolutely, yes.  The only normative statement required here is "MUST NOT reuse keys across origins".

martinthomson:
#130 is where the action is now.

Issue #94, opened 2014-12-10
unregister() should be available on the PushRegistration object


mounirlamouri:
I think it would be great to follow the pattern of ServiceWorker:
- unregister() should live in PushRegistration;
- unregister() should return a Promise<boolean> that says whether the unregistration happened.

CC: @annevk and @jakearchibald

mvano:
Agreed it would be good to align this with the sw spec, no need to do something different for a very similar feature. For reference, here's the sw unregister method definition: https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#navigator-service-worker-unregister

mounirlamouri:
Fixed by #97.

PR #97, opened 2014-12-15
Move unregister() to PushRegistration.

mounirlamouri:
This is also changing the type of returned Promise and how it is
expected to behave.

### DIFF
--- a/index.html
+++ b/index.html
@@ -171,9 +171,10 @@
         <code><a href=
         "http://www.w3.org/TR/dom/#notfounderror"><dfn>NotFoundError</dfn></a></code>,
         <code><a href=
-        "http://www.w3.org/TR/dom/#securityerror"><dfn>SecurityError</dfn></a></code>, and <a href=
-        "http://www.w3.org/TR/dom/#concept-event-constructor"><dfn>steps for constructing
-        events</dfn></a> are defined in [[!DOM]].
+        "http://www.w3.org/TR/dom/#securityerror"><dfn>SecurityError</dfn></a></code>,
+        <code><a href="http://www.w3.org/TR/dom/#networkerror"><dfn>NetworkError</dfn></a></code>,
+        and <a href="http://www.w3.org/TR/dom/#concept-event-constructor"><dfn>steps for
+        constructing events</dfn></a> are defined in [[!DOM]].
       </p>
       <p>
         <dfn>Service Worker</dfn>, <dfn>ServiceWorkerRegistration</dfn>,
@@ -442,9 +443,6 @@ navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) {
           Promise&lt;PushRegistration&gt; register ()
         </dt>
         <dt>
-          Promise&lt;PushRegistration&gt; unregister ()
-        </dt>
-        <dt>
           Promise&lt;PushRegistration&gt; getRegistration ()
         </dt>
         <dt>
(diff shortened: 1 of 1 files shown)


mounirlamouri:
Fixes issue #94.

index.html
```diff
@@ -662,6 +638,29 @@
...
+        "widl-PushRegistration-unregister-Promise-boolean">unregister</dfn></code> method when
+        invoked MUST run the following steps:
+      </p>
+      <ol>
+        <li>Let <var>promise</var> be a new <a><code>Promise</code></a>.
+        </li>
+        <li>Return <var>promise</var> and continue the following steps asynchronously.
+        </li>
+        <li>If the <a>webapp</a> is not registered, resolve <var>promise</var> with
+        <code>false</code> and terminate these steps.
```
mvano, line 652: I think we were still considering whether this should return true, right?

index.html
```diff
@@ -662,6 +638,29 @@
...
+        </li>
+        <li>Make a request to the system to deactivate the <a>push registration</a> associated with
+        the <a>webapp</a>.
+        </li>
+        <li>If it was not possible to access the <a>push server</a>, reject <var>promise</var> with
+        a "<code><a>NetworkError</a></code>" exception and terminate these steps.
+        </li>
+        <li>When the request has been completed, resolve <var>promise</var> with a
```
mvano: just with.

index.html
mounirlamouri, line 652: How so?

mvano, line 652: For the case where unregister is called twice. The desired result was achieved: you are unregistered. That sounds more like true than false.

mounirlamouri, line 652: I disagree. The entire idea of returning a boolean is to point that nothing actually happen. This is also what we do for ServiceWorker.

mvano, line 652: I think it's weird, but I don't feel too strongly about it and aligning with sw is good. lgtm

mounirlamouri:
Thanks! \o/

martinthomson:
Why do you need to know if there was something to unregister?  If there is a use case, I'd be interested in learning of it.

mounirlamouri:
Do we really need a use case to expose an information that can anyway be accessed by other means?

martinthomson:
Even more so then.  If you have other means of getting this information, I think that raises the bar.  Can you justify why this returns a `boolean` over `Promise<undefined>`?

mounirlamouri:
Because then you can easily check if the call was a no-op or not... If you have a `pushRegistration` instance, you can just call `pushRegistration.unregister().then(function(b) {...});` and use `b` to know whether the unregistration happened instead of doing `navigator.push.getRegistration().then(function(r) { return r ? r.unregister() : null; }).then(function() { ... });`.

martinthomson:
Why do you need to know if the call was a no-op?

mounirlamouri:
Should we wait for the need to arise before exposing that simple boolean? I'm not sure. Exposing this doesn't cost us anything and could help developers. Is there an implementation concern?

martinthomson:
we don't add API points without need usually, that's all

mvano:
If we later decided that we wanted the boolean, there would be some subtlety in the backward compatibility. Authors would need to distinguish between (false) and (undefined or zero arguments).

martinthomson:
Yes, this closes certain avenues if we later determine that information like this is useful.  That's not a reason to add features.

Issue #96, opened 2014-12-15
.getRegistration() should not reject if there is no registration (but return a null object)


mounirlamouri:
CC @jakearchibald

It seems to be the pattern generally being used with Promises. Rejecting should be an exceptional case. In that case, having no registration isn't exceptional, it's definitely expected.

Also, as @jakearchibald pointed, it would allow to do things like that in ES7:

``` javascript
var reg = await push.getRegistration();
```

mvano:
Do you have references to `the pattern generally being used with Promises`?

mvano:
Note that things would be a little different if we were talking about `Promise<bool> isRegistered()` which seems even less appropriate to throw an exception for.

jakearchibald:
http://www.w3.org/2001/tag/doc/promises-guide#rejections-should-be-exceptional

> What exactly you consider "exceptional" is up for debate, as always. But, you should always ask, before rejecting a promise: if this function was synchronous, would I expect a thrown exception under this circumstance? Or perhaps a failure value (like null, false, or undefined)?
> …
> Bad uses of rejections include:
> 
> When a value is asked for asynchronously and is not found: for example asyncMap.get("key") should return a promise for undefined when there is no entry for "key"

There's a big discussion over at https://github.com/slightlyoff/ServiceWorker/issues/374 where I fought for `caches.match` to reject if no match was found. I was talked around in the end.

mvano:
Thanks for the references Jake. I think those together with `await` make a reasonable argument for resolving with `null`.

martinthomson:
I assume that the absence of meta-information about the absence is addressed by querying some other API?  Like a [.permission](http://www.w3.org/TR/notifications/#dom-notification-permission) attribute perhaps?

mounirlamouri:
Not sure what you mean. I don't think `getRegistration()` should pass information about the absence of information. If the website never registered, it will have no registration, this is definitely expected. If the website tries to register but the call failed, `register()` should pass information about the failure reason, not `getRegistration()`. WDYT?

martinthomson:
If `getRegistration()` fails, one advantage of a rejection is that it gives you somewhere to hang the meta-information about that failure: is this because you are on an http:// origin?  Is this because the push server is down?  Is this because you haven't got consent?  Error codes determine what steps need to be taken to correct the error.

martinthomson:
I'm sorry, is this suggesting an asymmetry between `getRegistration()` and `register()`.

[...]

PR #101, opened 2015-01-06
Resolve with null if not registered.

mvano:
Issue #96


index.html
```diff
@@ -512,14 +510,12 @@
...
-        <li>If the <a>webapp</a> is not registered, reject <var>promise</var> with a
-        <code><a>DOMException</a></code> whose name is "<code><a>NotFoundError</a></code>" and
-        terminate these steps.
+        <li>If the <a>webapp</a> is not registered, resolve <var>promise</var> with null.
         </li>
         <li>Retrieve the <a>push registration</a> associated with the <a>webapp</a>.
         </li>
         <li>If there is an error, reject <var>promise</var> with a <code><a>DOMException</a></code>
-        whose name is "<code><a>AbortError</a></code>" and terminate these steps.
+        and terminate these steps.
```
beverloo: What kind of error is an implementation meant to throw if it's not an AbortError? Specifying the kind of DOMException still seems valuable here, even if you keep it as generic as AbortError.

beverloo:
lgtm % comment

mvano:
Thanks for taking a look guys. Let's just focus on making this thing nullable. The concepts of retrieving and the way to define errors should really be separate.

PR #106, opened 2015-01-13
Define PermissionDeniedError.

mvano:


index.html
```diff
@@ -874,7 +869,33 @@
...
+          </td>
+        </tr>
+      </table>
+    </section>
+    <section>
+      <h2>
+        Exceptions
+      </h2>
+      <p>
+        The Push API uses the following new <a>DOMException</a> names.
```
beverloo, line 882: Should we maintain a reference to the bug (maybe even as an issue) to avoid two specifications mentioning they're about to introduce the PermissionDeniedError?

mvano, line 882: Afaict it would not be a major problem to have a duplicate definition for a short time, see Anne's comment [1]. But of course we should clean it up once that situation arises. I filed an issue on the Push API spec for this: https://github.com/w3c/push-api/issues/107

[1] https://www.w3.org/Bugs/Public/show_bug.cgi?id=23033#c12

Issue #71, opened 2014-10-03
Multiple registrations per origin/SW


martinthomson:
I know that this was talked about previously, but I'd like to challenge that decision.  

(I'm not clear whether the restriction is per origin or per service worker scope.  I don't think that matters here.)

I have several reasons:
1. Origins can have multiple applications.  Having them share a push notification registration is a significant burden on those applications.  And even within those applications, separation of functions can be valuable.  For example, an application as simple as gmail [sic] benefits from different notification channels for email, chat messages and calling, not to mention software update and a host of other things.  Sure, the application can perform demux functions, but that creates a need for a centralized dispatcher: a function that the browser is performing anyway.
2. We have a request to support what GCM calls a collapse key.  A function whereby only the last messages with the same key is delivered if a device is absent.  This is much easier with the proposed protocol if multiple registrations (there, channels) can be created (it's basically a noop).
3. This has better privacy and security properties.  A collapse key provides a way for the push server to identify correlation between unrelated events, which is unavoidable if you have the feature.  However, with a special collapse key in addition to the registration, the push server is able to correlate un-collapsed messages for the same origin.  Separate channels might still be activated by the same application server, revealing a correlation through the source IP of the push request, but that certainly isn't always the case.  This also allows applications to more finely control disclosure of registration information.

mvano:
In the current design, there can be one PushRegistration per ServiceWorkerRegistration.

1: Origins can have multiple applications, but usually they have different paths. Service worker scopes handle this. I see your point about having a centralized dispatcher inside a complex application. However, I don't see that as a big hurdle, given the complexity of the application I'm sure the engineers can handle it. They might want such a dispatcher anyway for other coordination reasons. I'm not yet convinced this requires a more complex API.
1. I'm not sure what you mean here. Collapse keys currently work just fine with GCM, which also use a single push registration.
2. Doesn't the problem of correlation go away with encryption? Or does it weaken encryption? Anyway, there are other, very explicit, signals for correlation in GCM  such as sender_id, application_id, and api_key. This is for good reason, they are used to combat abuse like spam or DOS attacks. What system are you thinking of that does not do that? How does it fight abuse?

martinthomson:
OK, that numbering is very confusing...

[...]

Issue #53, opened 2014-09-30
Clean up "webapp server" terminology


mvano:
We currently use a few variant terms for the same thing:
- Webapp server
- App server
- Application server

We should try to agree with the IETF Web Push Protocol WG on picking one and using only that one.

beverloo:
I think "webapp server" is too narrow, given that the webapp may just be the front-end of a much larger application.

Since "app" is commonly used for client-side things, whether it be installable native apps or web apps, my preference has "application server".

mvano:
See latest IETF draft: http://tools.ietf.org/html/draft-thomson-webpush-http2-00

efullea:
I support using "application server"

mvano:
Where does that leave the term "webapp" then? Should it be:
- Webapp
- App
- Application

efullea:
I think 'webapp' is fine in this case as AFAIU it is the scope of this W3C spec.

beverloo:
Agreed.

martinthomson:
SGTM.  As long as we consider 'webapp' to be resident on the 'device', that works out.

I don't want to start using webapp except in providing context, because I think that the protocol aspects of this could go beyond the web.

PR #112, opened 2015-02-03
Clarify subscription relationships, lifecycle, and uniqueness.

mvano:
Terminology and concepts:
- Moves up and clarifies some crucial details about the number of subscriptions, their associations, and their lifecycles. These were spread out further down the document in various unrelated sections.

For the section for Security and privacy considerations:
- Moves some relevant details about subscriptions into this section
- Calls out explicitly some considerations about uniqueness that might otherwise be implicit, and adds some examples of why this matters.

For number of subscriptions:
- This change relates to issue #71 as it removes an inconsistency around the number of subscriptions. Before this change, one part of the spec indicated there was one subscription per webapp, but another indicated there was one per service worker registration. The least desirable case, as discussed in issue #71 (one per webapp) is removed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -217,56 +217,91 @@
     </section>
     <section>
       <h2>
-        Terminology
+        Concepts
       </h2>
-      <p>
-        The term <dfn>webapp</dfn> refers to a Web application, i.e. an application implemented
-        using Web technologies, and executing within the context of a Web <a>user agent</a>, e.g. a
-        Web browser or other Web runtime environment.
-      </p>
-      <p>
-        The term <dfn>webapp server</dfn> refers to server-side components of a <a>webapp</a>.
-      </p>
-      <p>
-        A <dfn>push message</dfn> is an indication to a <a>webapp</a> that there is new information
-        for it from the <a>webapp server</a>, all or part of which MAY be contained in the <a>push
-        message</a> itself.
-      </p>
-      <p>
-        A <dfn>push subscription</dfn> is a message delivery context that is established between
-        the <a>user agent</a> and the <a>push service</a> on behalf of a <a>service worker
-        registration</a>, and it is associated with that <a>service worker registration</a>. A
-        <a>push subscription</a> MUST NOT persist longer than its associated <a>service worker
-        registration</a>, but it MAY be terminated earlier.
-      </p>
-      <p>
-        A <a>push subscription</a> has an associated <dfn>endpoint</dfn>. It MUST be the absolute
-        URL exposed by the <a>push server</a> where the <a>webapp server</a> can send <a title=
-        "push message">push messages</a>. It MAY be the same for all <a title=
-        "push subscription">push subscriptions</a> associated with a <a>push service</a>.
-      </p>
-      <p>
-        A <a>push subscription</a> has an associated <dfn>subscription id</dfn>. It is used by the
-        <a>webapp server</a> to indicate the target of the <a title="push message">push
-        messages</a> that it submits to the <a>push server</a>. Each pair of <a>subscription id</a>
-        and <a>endpoint</a> is expected to be globally unique.
-      </p>
-      <p>
-        The term <dfn>push service</dfn> refers to an overall end-to-end system that allows
-        <a title="webapp server">webapp servers</a> to send <a title="push message">push
-        messages</a> to a <a>webapp</a>.
-      </p>
-      <p>
-        The term <dfn>push server</dfn> refers to the <a>push service</a> access point via which
-        <a title="webapp server">webapp-servers</a> can initiate <a>push message</a> delivery. Push
-        servers typically expose APIs specific to the <a>push service</a>, e.g. for <a>push
-        message</a> delivery initiation.
-      </p>
-      <p>
-        The term <dfn>express permission</dfn> refers to an act by the user, e.g. via user
-        interface or host device platform features, via which the user approves the permission of a
-        <a>webapp</a> to access the Push API.
-      </p>
+      <section>
+        <h2>
+          Application
+        </h2>
+        <p>
+          The term <dfn>webapp</dfn> refers to a Web application, i.e. an application implemented
+          using Web technologies, and executing within the context of a Web <a>user agent</a>, e.g.
+          a Web browser or other Web runtime environment.
+        </p>
+        <p>
+          The term <dfn>webapp server</dfn> refers to server-side components of a <a>webapp</a>.
+        </p>
+      </section>
+      <section>
+        <h2>
+          Push message
+        </h2>
+        <p>
+          A <dfn>push message</dfn> is an indication to a <a>webapp</a> that there is new
+          information for it from the <a>webapp server</a>. All or part of this information MAY be
+          contained in the <a>push message</a> itself.
+        </p>
+        <p>
+          A <a>push message</a> MUST be delivered to the <a>active worker</a> associated with the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -223,14 +228,28 @@
         The term <dfn>webapp server</dfn> refers to server-side components of a <a>webapp</a>.
       </p>
       <p>
-        The term <dfn>push message</dfn> refers to an indication to a <a>webapp</a> that there is
-        new information for it from the <a>webapp server</a>, all or part of which MAY be contained
-        in the <a>push message</a> itself.
+        A <dfn>push message</dfn> is an indication to a <a>webapp</a> that there is new information
+        for it from the <a>webapp server</a>, all or part of which MAY be contained in the <a>push
```
beverloo: The protocol calls this "application server", as does your code comment. It would be good to be consistent there.

index.html
```diff
@@ -223,14 +228,28 @@
...
+        message</a> itself.
+      </p>
+      <p>
+        A <dfn>push subscription</dfn> is a message delivery context that is established between
+        the <a>user agent</a> and the <a>push service</a> on behalf of a <a>service worker
+        registration</a>, and it is associated with that <a>service worker registration</a>. A
+        <a>push subscription</a> MUST NOT persist longer than its associated <a>service worker
+        registration</a>, but it MAY be terminated earlier.
```
beverloo: What about this:

A push subscription is a message delivery context established between the user agent and the push service on behalf of a webapp, associated with the webapp's service worker registration.

It avoids the double "service worker registration" term, s/it// and s/that is//.

index.html
```diff
@@ -223,14 +228,28 @@
...
+      </p>
+      <p>
+        A <a>push subscription</a> has an associated <dfn>endpoint</dfn>. It MUST be the absolute
+        URL exposed by the <a>push server</a> where the <a>webapp server</a> can send <a title=
+        "push message">push messages</a>. It MAY be the same for all <a title=
```
beverloo: +to (after "push messaging"). This is strong language separating the endpoint from the subscriptionId, is this in line with the open issue about merging them?

index.html
```diff
@@ -223,14 +228,28 @@
...
+        "push subscription">push subscriptions</a> associated with a <a>push service</a>.
       </p>
       <p>
-        The term <dfn>push subscription</dfn> refers to each logical channel aimed at delivering
-        <a title="push message">push messages</a> from an <a>webapp server</a> to a <a>webapp</a>,
-        which is created by a distinct subscription of such <a>webapp</a> via this Push API.
+        A <a>push subscription</a> has an associated <dfn>subscription id</dfn>. It is used by the
+        <a>webapp server</a> to indicate the target of the <a title="push message">push
+        messages</a> that it submits to the <a>push server</a>. Each pair of <a>subscription id</a>
+        and <a>endpoint</a> is expected to be globally unique.
```
beverloo: I'd prefer to use normative terminology instead of "is expected". Is this a SHOULD or a MUST?

index.html
```diff
@@ -257,12 +276,21 @@
         <a title="user agent">User agents</a> MUST NOT provide Push API access to <a title=
         "webapp">webapps</a> without the <a>express permission</a> of the user. <a title=
         "user agent">User agents</a> MUST acquire consent for permission through a user interface
-        for each call to the <code>subscribe()</code> method, unless a prearranged trust
-        relationship applies.
+        for each call to the <code>subscribe()</code> method, unless a previous permission grant
+        has been persisted, or a prearranged trust relationship applies. Permissions that are
```
beverloo, line 315: Wouldn't we consider previously granted permissions to be a prearranged trust relationship in context of the current session?

index.html
```diff
@@ -257,12 +276,21 @@
...
+        preserved beyond the current browsing session MUST be revocable.
       </p>
       <p>
-        <a title="user agent">User agents</a> MAY support prearranged trust relationships that do
-        not require such per-request user interfaces.
+        When a permission is revoked, all <a title="push subscription">push subscriptions</a>
+        created with that permission MUST be terminated.
```
beverloo: s/terminated/unsubscribed/

index.html
```diff
@@ -257,12 +276,21 @@
...
+      </p>
+      <p>
+        When a <a>service worker registration</a> is unregistered, any associated <a>push
+        subscription</a> MUST be terminated.
```
beverloo: Dito.

index.html
```diff
@@ -257,12 +276,21 @@
...
+      </p>
+      <p>
+        The <a>subscription id</a> of a terminated <a>push subscription</a> MUST never be reused
+        for a new <a>push subscription</a>.
```
beverloo: Perhaps concisely clarify _why_ this is the case? (Persistent identifiers.)

I imagine that an implementer reading the specification will initially consider the semantics of the push service that will service their implementation. If that service generates a subscription Id based on some kind of hardware Id, it's not unrealistic to expect naive re-use. Hence why giving some context may be useful.

index.html
```diff
@@ -450,9 +475,8 @@
         <a>PushManager</a> interface
       </h2>
       <p>
-        The <a>PushManager</a> interface defines the operations that enable <a title=
-        "webapp">webapps</a> to establish access to <a title="push service">push services</a>. Note
-        that just a single <a>push subscription</a> is allowed per <a>webapp</a>.
```
beverloo, line 455: I can't seem to find where you reformulate the note about having a single subscription per webapp in the diff. Did I miss it? If it's a deliberate change, we shouldn't include it in this PR.

index.html
mvano: This is issue #53, currently the defined term is called webapp server. Changing this should be part of addressing issue #53.

index.html

index.html
mvano: This is a slightly rephrased section that used to accompany the IDL attribute description. I feel the rephrasing made it clearer and more correct. This change has nothing to do with the capability url discussion though. It seems both sides of that discussion can still conform to this section, due to the use of MAY, and it seems inter-operable as well.

index.html
mvano: I think this can be a MUST. If there are reasonable exceptions that indicate the need for a SHOULD, let's have a separate issue with a discussion for that.

index.html
mvano, line 315: A prearranged trust relationship can also be part of an EULA or something like that. The use of persisted permission grants is by far the most common thing for web APIs, so it makes sense to call it out, for clarity. Finally, whether a persisted grant is "prearranged" seems open to interpretation. The new phrasing seems clearer to me.

index.html
mvano: Rephrased.

index.html
mvano, line 455: This was an inconsistency in the spec and its removal is intentional. I've updated the change description to clarify.

index.html
mvano: Went for deactivated.

index.html
mvano: Same.

index.html
```diff
@@ -214,40 +214,94 @@
...
-        <a>webapp</a> to access the Push API.
-      </p>
+    </section>
+    <section>
+      <h2>
+        Concepts
+      </h2>
+      <section>
+        <h2>
+          Application
```
beverloo: The title is application, but the definition talks about "webapp". I understand that you want to move away from "webapp", but let's not make this more inconsistent than it needs to be?

beverloo:
Ok % comment.

PR #124, opened 2015-03-11
Removed NoInterfaceObject from PushMessageData (closes #123)

### LINKED ISSUE #123: Remove [NoInterfaceObject] from PushMessageData
Please stop using [NoInterfaceObject]. It's bad for extensibility. For instance, if we I want to polyfill a new processing method on `PushMessageData` (i.e., .fancyMsgFormat()), I won't be able to do that without it being a huge pain the ass.

marcoscaceres:


mounirlamouri:
lgtm. I will let the editors do the merge though.

johnmellor:
The reasoning was that we might want to later standardize in WebIDL the concept of a bunch of binary data that can be consumed as arrayBuffer/blob/text/json, and if that ends up having a superset of this PushMessageData, we would be able to update the spec to use that without breaking backwards compatibility.

But looking back over https://github.com/w3c/push-api/pull/88, I realise that we already diverge from the Fetch spec's [similar concept](https://fetch.spec.whatwg.org/#body) (PushMessageData can be read multiple times, and can't currently be consumed as FormData). So maybe this is fine.

PR #126, opened 2015-03-17
Fixing definition of push message

martinthomson:
The definition here was a little strange, probably as a legacy of the original "no data" API.

It also contained unnecessary normative language.

### DIFF
--- a/index.html
+++ b/index.html
@@ -238,14 +238,14 @@
           Push message
         </h2>
         <p>
-          A <dfn>push message</dfn> is an indication to a <a>webapp</a> that there is new
-          information for it from the <a>application server</a>. All or part of this information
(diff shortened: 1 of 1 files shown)


beverloo:
LGTM.

Michael is currently out of office for a few weeks. Perhaps @efullea is around?

martinthomson:
> [...] it doesn't properly describe push messages with no body [...]

Yeah, the degenerate case is ugly, but I think that's OK.

As for timing, I expect to generate a few more, some of which might be controversial.  We can instruct Michael how to proceed if we can agree.  Or we can defer the discussion.  I don't necessarily feel the same amount of rush as you do.

mvano:
Nice, much more elegant. I think there is a ton more text in the spec that can be streamlined like this.

PR #129, opened 2015-03-17
Move to using just endpoint

### LINKED ISSUE #56: Remove registrationId
Currently, a registration looks like this:

``` webidl
interface PushRegistration {
    readonly    attribute DOMString endpoint;
    readonly    attribute DOMString registrationId;
};
```

Which is two values, which are used in two ways:
1. both together, they are used to send a message
2. `registrationId` is used to identify a registration in the API

Requiring two identifiers is a burden on the application when it comes to sending messages.  The proposed web push protocol needs a single URI only.  The push server can embed any information it needs in the URI.

Using the same identification that the push server uses in the API is also unnecessary.  Other APIs, if they have some sort of identification, use identifiers that are minted by the browser.  Given that we (currently) have only a single URI per origin, we don't need to perform this identification at all.

I'd like to propose the removal of `registrationId`.

martinthomson:
There was a lot of text here that was simply wrong, probably because it was outdated.

I've removed or reworked that part.  I haven't cited the IETF work (yet), but there is a placeholder there.

Closes #56.


martinthomson:
Oh yeah, I forgot, this doesn't update the big sequence diagram.  I might get to that, but without the original source, I'd have to start over.

beverloo:
I agree that we need to make a decision about this, while I'm still erring towards the separated properties side myself. The problem is that no one feels strongly about it - there's arguments both for keeping it and for removing it.

Let's wait for @mvano's return and then find a way to drive this to a solution.

index.html
```diff
@@ -418,71 +417,21 @@
       </section>
       <section>
         <h3>
-          Push Service Discovery and API Use
+          Push API Use
```
mvano: I think you lost some meaningful information in this heading. How about something like "Push service use" as what we really care about here is pointing to the IETF Web Push Protocol.

index.html
```diff
@@ -418,71 +417,21 @@
...
-          </li>
-        </ul>
-        <p>
-          Push services that are compatible with the Push API will be expected to clarify how these
-          data are used in their APIs, e.g. through their developer programs. Examples of
-          considerations include:
+          The fields included in the <code>PushSubscription</code> is all the information needed for
+          an <a>application server</a> to send a <a>push message</a>. Push services that are
+          compatible with the Push API will be expected to provide an <a>endpoint</a> that conforms
+          to <em>TODO: refer to IETF spec</em>.  These parameters and attributes include:
```
mvano: Can this TODO be resolved now?

index.html
```diff
@@ -418,71 +417,21 @@
...
-          or <code>subscriptionId</code> be present in some body element.
-          </li>
-          <li>The <code>PushSubscription</code> attributes may only be used by the <a>application
-          server</a> to associate the webapp to a specific push service subscription, with the
-          details of the push service API relying upon other pre-configured data.
+          <li>The <code><a>endpoint</a></code> of a <code>PushSubscription</code> is a URI that
+          allows an <a>application server</a> to request delivery of a <a>push message</a> to
+          a <a>webapp</a> subscription.
           </li>
+          <li><em>TODO: More fields to add.</em></li>
```
mvano: This item is not very useful. Please file specific issues for feature requests or bugs. No need to refer to those issues here either, unless they help resolve inconsistencies internal to this document.

index.html
```diff
@@ -418,71 +417,21 @@
...
-          flexible means for push services to deliver webapp-specific subscription and/or push
-          service API parameters to the <a>application server</a>, through the webapp.
-          </li>
-        </ul>
-        <p>
-          Push services that are compatible with the Push API will be expected to clarify how these
-          data are used in their APIs, e.g. through their developer programs. Examples of
-          considerations include:
+          The fields included in the <code>PushSubscription</code> is all the information needed for
+          an <a>application server</a> to send a <a>push message</a>. Push services that are
```
mvano: How about working in something like this: "Push services SHOULD support the IETF Web Push Protocol" with a link to _something_. Add a note indicating that it's a work in progress.

index.html
```diff
@@ -418,71 +417,21 @@
...
+          compatible with the Push API will be expected to provide an <a>endpoint</a> that conforms
```
mvano: Detailed definition of the endpoint should really go in the concepts section, or the dependencies section if it is defined in the IETF spec.

mvano:
Thanks for the simplifications Martin! Now that the IETF work is progressing and has support from Mozilla, Microsoft, Google, and interest from others as well, this generally looks good to me. If you could address some of my comments, this pull request should get merged ASAP.

index.html
```diff
@@ -408,6 +404,9 @@
         <h2>
           Sequence diagram
         </h2>
+        <p class="issue">
+          This needs updating and there is no source in the repo (naughty).
```
mvano: I've added the sequence diagram source in vdx format. I made it in Lucidchart, and editing it should work well there.

index.html
martinthomson: Wait a little longer.  We need people to review and comment on the proposed protocol; and to express their view that this should be adopted.

index.html
martinthomson: I wanted to avoid having to create a biblio entry for a proposal.  As above, I hope that the IETF WG will adopt this soon.

martinthomson:
OK, I just had a shot at this.  I made some fairly big changes to the diagram, so check those over.  I don't know how you want to do comments; maybe I can share it in Lucidchart.

paul-em:
I think it is way better to understand. 

The only minor things I found: Shouldn't clear push subscription happen after unsubscription, so that you can wait for it to resolve successfully before you delete it? 
And I find "set push subscription" not very informative, why not take the old "store push subscription" or "save push subscription", or was there another thought behind?

martinthomson:
@paul-em, you shouldn't remove a subscription before you tell everyone that is using it to stop using it.  Otherwise, they will expect to have messages delivered to you and (probably) get errors.

set vs save vs store matters not to me (other than the time it will now take to regenerate all the files).  I thought that set and clear were neatly symmetrical, but didn't really think that hard about it.

paul-em:
Oh ok makes sense to remove before then.

mvano:
Martin, the diagram changes generally look good to me. Just a few nits:
1) the clear push subscription should have a PushSubscription as an argument, like the set action
2) If you can crop the image and drop the blank bits on the edges, less screen space will be wasted. We can always add padding in CSS if needed.

mvano:
Hmm, maybe you can address those nits later, let's just get this merged for now.

PR #130, opened 2015-03-18
Message encryption

martinthomson:
This builds on #129 by adding encryption based on https://tools.ietf.org/html/draft-thomson-webpush-encryption.  At the moment, this is just a strawman for discussion.

### DIFF
--- a/index.html
+++ b/index.html
@@ -704,12 +704,12 @@ navigator.serviceWorker.register('serviceworker.js').then(
         </dt>
       </dl>
       <p>
-        When getting the <code><dfn data-lt="PushSubscription-endpoint">endpoint</dfn></code>
+        When getting the <code><dfn id="widl-PushSubscription-endpoint" data-lt="PushSubscription-endpoint">endpoint</dfn></code>
         attribute, the <a>user agent</a> MUST return the <a>endpoint</a> associated with the
         <a>push subscription</a>.
       </p>
       <p>
-        The <code><dfn data-lt="PushSubscription-getKey">getKey</dfn></code> method retrieves an
+        The <code><dfn id="widl-PushSubscription-getKey-ArrayBuffer-PushEncryptionKeyName-name" data-lt="PushSubscription-getKey">getKey</dfn></code> method retrieves an
         encryption key using the following process:
       </p>
       <ol>
--- a/index.html
+++ b/index.html
@@ -492,10 +492,10 @@ navigator.serviceWorker.register('serviceworker.js').then(
           </li>
           <li>The <code><a data-lt="PushSubscription-getKey">getKey</a></code> method on a
           <code>PushSubscription</code> is used to retrieve keying material used to encipher
-          <a>push messages</a>. Each invocation of the function returns a
-          new <code>ArrayBuffer</code> that contains the value of the corresponding key,
-          or <code>null</code> if the identified key doesn't exist. Passing a value
-          of <code>p256dh</code> retrieves a <dfn data-lt= "ECDH">elliptic curve Diffie-Hellman
+          <a>push messages</a>. Each invocation of the function returns a new
+          <code>ArrayBuffer</code> that contains the value of the corresponding key, or
+          <code>null</code> if the identified key doesn't exist. Passing a value of
+          <code>p256dh</code> retrieves a <dfn data-lt="ECDH">elliptic curve Diffie-Hellman
           (ECDH)</dfn> public key associated with the <a>push subscription</a>. This key is used by
           the <a>application server</a> to encipher messages for the <a>push subscription</a>, as
           described in [[!WEBPUSH-ENCRYPTION]].
@@ -731,9 +731,9 @@ navigator.serviceWorker.register('serviceworker.js').then(
         instance.
         </li>
         <li>Set the contents of <var>key</var> to the serialized value of the public key from the
-        key pair. This uses the serialization format described in the specification that defines the
-        name. For example, [[!WEBPUSH-ENCRYPTION]] specifies that the <code>p256dh</code> public key
-        is encoded using the uncompressed format defined in [[X9.62]] Annex A.
+        key pair. This uses the serialization format described in the specification that defines
+        the name. For example, [[!WEBPUSH-ENCRYPTION]] specifies that the <code>p256dh</code>
+        public key is encoded using the uncompressed format defined in [[X9.62]] Annex A.
         </li>
         <li>Return <var>key</var>.
         </li>
--- a/index.html
+++ b/index.html
@@ -694,7 +694,7 @@ navigator.serviceWorker.register('serviceworker.js').then(
           readonly attribute USVString endpoint
         </dt>
         <dt>
-          readonly attribute ArrayBuffer? getKey (PushEncryptionKeyName name)
+          ArrayBuffer? getKey (PushEncryptionKeyName name)
         </dt>
         <dt>
           Promise&lt;boolean&gt; unsubscribe ()
--- a/index.html
+++ b/index.html
@@ -433,6 +433,7 @@ navigator.serviceWorker.register('serviceworker.js').then(
     serviceWorkerRegistration.pushManager.subscribe().then(
       function(pushSubscription) {
         console.log(pushSubscription.endpoint);
+        console.log(pushSubscription.p256dh);
         // The push subscription details needed by the application
         // server are now available, and can be sent to it using,
         // for example, an XMLHttpRequest.
@@ -465,14 +466,26 @@ navigator.serviceWorker.register('serviceworker.js').then(
           Push service use
         </h3>
         <p>
-          A <code>PushSubscription</code> contains all the information needed to send a <a>push
-          message</a>. A <a>push message</a> can be delivered using the <a>web push protocol</a>.
+          The fields included in the <code>PushSubscription</code> is all the information needed for
+          an <a>application server</a> to send a <a>push message</a>. Push services that are
+          compatible with the Push API will be expected to provide an <a>endpoint</a> that conforms
+          to the <a>web push protocol</a>.  These parameters and attributes include:
         </p>
-        <p>
-          The <code><a>endpoint</a></code> of a <code>PushSubscription</code> is a URI that allows
-          an <a>application server</a> to request delivery of a <a>push message</a> to a <a>push
-          subscription</a>.
-        </p>
+        <ul>
+          <li>
+            The <code><a>endpoint</a></code> of a <code>PushSubscription</code> is a URI that allows
+            an <a>application server</a> to request delivery of a <a>push message</a> to
+            a <a>webapp</a>.
+          </li>
+          <li>
+            The <code>p256dh</code> of a <code>PushSubscription</code> contains the P-256 public
+            <dfn id="ECDH">elliptic curve Diffie-Hellman (ECDH)</dfn> key associated with
+            the <a>push subscription</a>.  This key is used by the <a>application server</a> to
+            encipher messages for the <a>push subscription</a>, as described
+            in <a href="https://tools.ietf.org/html/draft-thomson-webpush-encryption">TODO cite the
+            spec</a>.
+          </li>
+        </ul>
       </section>
     </section>
     <section>
@@ -576,6 +589,10 @@ navigator.serviceWorker.register('serviceworker.js').then(
         <li>If there is an error, reject <var>promise</var> with a <code><a>DOMException</a></code>
         whose name is "<code><a>AbortError</a></code>" and terminate these steps.
         </li>
+        <li>Generate a new P-256 <a>ECDH</a> key pair.  Store the private key in an internal slot
+        associated with the subscription; this value MUST NOT be made available to applications.
+        Store the public key in the <code>p256dh</code> field of the subscription.
+        </li>
         <li>When the request has been completed, resolve <var>promise</var> with a
         <a><code>PushSubscription</code></a> providing the details of the new <a>push
         subscription</a>.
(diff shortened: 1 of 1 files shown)


beverloo:
@martinthomson - Was there agreement at the IETF about the specific curve to use (P-256 / 25519 I believe)? As I understand it, AGL prefers the latter, but Costin had concerns about deployability as it's still somewhat new.

+@fgorski

martinthomson:
I talked to agl about this and he asked me about this specific point.  If 25519 were more widely deployed, I'd definitely use it (it uses fewer bits to express a public key, it's faster, etc...).  However, P-256 enjoys far better deployment.  For instance, 25519 requires a lot of extra work to implement in Firefox, but we've had P-256 for years.

@costinm original concerns were around deployment to old versions of Android, but old versions of Android lack both AES-GCM and P-256 as well as 25519.  I don't know whether there are any versions with 25519, but it is likely less well-deployed than P-256.

BTW, once Brian and I get a revised draft submitted, I will submit a revision to this PR with proper links.

martinthomson:
Updated with the latest.

index.html
```diff
@@ -820,10 +841,24 @@
...
+          with <var>subscription</var> and the process described in <a href=
+          "https://tools.ietf.org/html/draft-thomson-webpush-encryption">TODO: cite</a>. This
+          produces the plain text of the message.
+          </li>
+          <li>If the <a>push message</a> could not be decrypted for any reason, or if it is not
+          encrypted, discard the message and terminate this process. A <code>push</code> event MUST
+          NOT be fired for a <a>push message</a> that was not successfully decrypted using the key
+          pair associated with the <a>push subscription</a>.
+          </li>
+          <li>Let <var>pushdata</var> be the decrypted plain text of the <a>push message</a>.
```
beverloo, line 968: Using the term "plain text" makes sense from a cryptographic point of view, but it could confuse Web developers to whom the definition of "plain text" is analogous to `Content-Type: text/plain`.

I'm not proposing a change, but do want to raise this as a possible pitfall.

beverloo:
Overall SGTM, pending the on-going discussion about the curve to use. We're experimenting with `curve25519dh` following the semantics of the PR.

Something to consider for a follow-up: giving authors the ability to explicitly rotate their keys, e.g. `PushSubscription.prototype.rotateKeys()`. This would allow them to avoid creating an entirely new subscription in case it's just the keys they'd like to change.

martinthomson:
I'm not sure that rotating is going to be worth the effort.  Rotating keys means that you have to signal the new keys to the application server; and ending the use of the old keys means that you have effectively made the subscription unusable anyway.

beverloo:
That depends on how we plan to solve broadcast messages - if these are allowed either without encryption, or, more likely, without payload, this would only impact part of the use.

Nonetheless, let's defer until we learn about concrete use-cases :).

martinthomson:
In review at Mozilla, Olli noted that the strange and cryptic "p256dh" attribute might be better replaced with:

```
enum PushSubscriptionKeyType { "p256dh" };
partial interface PushSubscription {
  ArrayBuffer? getEncryptionData(PushSubscriptionKeyType type);
};
```

@beverloo, do you have any opinion there?  I think that this is a good suggestion.

beverloo:
It definitely makes the code easier to read for people unaware of the encryption semantics, so +1.

Could we use `getEncryptionKey` rather than `getEncryptionData`? _data_ could be anything.

martinthomson:
Key is fine.

Hmm, thinking about this some more, we would need to describe a new serializer.

annevk:
Bikeshed: `getKey()`. No need for long names when usage of the method and its single argument make everything clear.

martinthomson:
That's a great colour for the bike shed.

beverloo:
Agreed.

Having a method raises an interesting question for the serializer - it'd expose a property that otherwise wouldn't exist. However, since it's not a client-side concern there's a good reason for `JSON.stringify(pushSubscription)` to include anything the server needs.

It could either include the escaped hex sequence of the ArrayBuffer's contents, or the base64 representation. I'm not sure what would be idiomatic - there's little precedence.

costinm:
getPublicKey ? Is the parameter optional ( since enum can have only one
possible value anyways ) ?

Also it should document very clearly what's in the ArrayBuffer - the DER
encoding ? And how do we expect
apps to send it to the app server ?

martinthomson:
I've updated the PR, which I think is largely OK.

@costinm:

> getPublicKey ? Is the parameter optional ( since enum can have only one
> possible value anyways ) ?

I don't see much advantage in making the parameter optional.  On the syntactic noise vs. being specific debate I always fall in the latter camp.

> Also it should document very clearly what's in the ArrayBuffer - the DER
> encoding ? And how do we expect apps to send it to the app server ?

The specification describes the encoding that is expected: https://tools.ietf.org/html/draft-thomson-webpush-encryption-01#section-5

Apps should be able to use the serializer on `PushSubscription`, that is, `JSON.stringify`.  The only question I have here is how `ArrayBuffer` attributes are handled.

martinthomson:
Hmm, I just looked into the situation with serializing `ArrayBuffer`.  It looks like the defined serialization is `{}`, which is no good.  If we are exporting keys, we need to ensure that we have a constant time string conversion for ArrayBuffer.

index.html
```diff
@@ -662,17 +689,45 @@
...
+        <li>Find the internal slot that contains the encryption key named by the <code>name</code>
+        argument.
+        </li>
+        <li>If an encryption key was not found, return <code>null</code>.
+        </li>
+        <li>If the encryption key represents an asymmetric key pair, set <var>key</var> to the
+        serialized value of the public key from the key pair. This uses the serialization format
+        defined in the specification for the name.
+        </li>
+        <li>If the encryption key represents a symmetric key, set <var>key</var> to the serialized
```
beverloo: No need to define behavior for symmetric keys until the enumeration lists one.

index.html
```diff
@@ -696,6 +751,57 @@
...
+            p256dh
+          </dt>
+        </dl>
+        <p>
+          The <code><dfn id="widl-PushEncryptionKeyName-p256dh">p256dh</dfn></code> value is used
+          to retrieve the P-256 ECDH Diffie-Hellman public key described in
+          [[!WEBPUSH-ENCRYPTION]].
+        </p>
+        <p>
+          Other encryption keys could be provided to applications.
```
beverloo: Do you explicitly want to allow for UA-specific extensions to the PushEncryptionKeyName values? (Why?)

WebIDL enums validate their values. As such, calling `getKey("petercurve9001")` would throw a TypeError exception rather than hit the "not-found" clause in the getKey algorithm.

index.html
```diff
@@ -696,6 +751,57 @@
...
+        </li>
+        <li>Let <var>keys</var> be an empty map.
+        </li>
+        <li>For each identifier <var>i</var> corresponding to keys in internal slots on the
+        <a><code>PushSubscription</code></a>, ordered by the name of the key:
+          <ol>
+            <li>Add an entry to <var>keys</var> whose key name is the name of the key and whose
+            value is an <code>ArrayBuffer</code> containing the defined serialization of the public
+            key (for an asymmetric key pair) or the key (for a symmetric key).
+            </li>
```
beverloo, line 794: What does this mean in practice? Given key `K` which is 32 octets of binary data, could you give a JSON example of the output you expect to retrieve?

index.html
```diff
@@ -426,6 +435,7 @@
     serviceWorkerRegistration.pushManager.subscribe().then(
       function(pushSubscription) {
         console.log(pushSubscription.endpoint);
+        console.log(pushSubscription.p256dh);
```
ddragana: this should reflect new way of getting key?

beverloo:
> If we are exporting keys, we need to ensure that we have a constant time string conversion for ArrayBuffer.

Excuse my naitivity, but why does this matter? The client-side already has access to the raw key using `getKey`, and given that the key is bound to the origin that shouldn't be timing sensitive. The server-side can thus receive the key in either format. Since we rely on the client-server connection being secure for transporting the key (HTTPS), does the timing sensitivity occur if the server were to store the key in, say, base64?

If the concern is that the client-server connection could be intercepted, then this problem could be attributed to the public key having to stay private more than timing sensitivity of serialization, right?

index.html
martinthomson: I'll remove it.

index.html
martinthomson, line 794: If `K` is 32 octets of binary data, then this is easy, but keys are often other than an octet string.  An elliptic curve public key is sometimes a pair of numbers, which then have to be compressed and encoded, or just encoded (where you define whether it is big- or little-endian).  This probably needs to refer to the encryption spec, which specifies the format for the public key we are using.

index.html
martinthomson: I wanted to leave this as an option for you specifically.  If you want to use curve 25519, then you could.  If you don't want to, then I'm happy to remove it.

martinthomson:
> > If we are exporting keys, we need to ensure that we have a constant time string conversion for ArrayBuffer.
> 
> Excuse my naitivity, but why does this matter? The client-side already has access to the raw key using getKey, and given that the key is bound to the origin that shouldn't be timing sensitive. The server-side can thus receive the key in either format. Since we rely on the client-server connection being secure for transporting the key (HTTPS), does the timing sensitivity occur if the server were to store the key in, say, base64?
> 
> If the concern is that the client-server connection could be intercepted, then this problem could be attributed to the public key having to stay private more than timing sensitivity of serialization, right?

I've been planning to more precisely articulate the concerns that your colleagues raised, which are:
1. Authentication of the application server relies on the application (as a whole) maintaining the confidentiality of one of the endpoint URL or the public key.  Leak both of these, and messages can be sent to the user agent.  If you leak one, the endpoint is perhaps more serious, because it can allow an attacker to send messages to the endpoint, which will drain battery, even if they aren't delivered.
2. These two values are not generally treated as secrets.

A URL is - in many contexts - leaked all over the place.  Even if we don't consider things like Referer to be relevant in this context, we have to concern ourselves with the potential for non-constant-time encoding and decoding of strings.  For example, translation between UTF-16 and UTF-8 happens all over the place in a browser and if that translation leaks bits of the URL into a timing side channel, it could be bad.  Of course, I'm not aware of any timing channel that reveals any difference between the usual base64 URL-safe alphabet characters, but my understanding is that this is primarily a principled position rather than a concern based on a specific attack.

Similarly, a public key is not treated with the same care as a private key.  Therefore, again, there might be leaks.  In this case, a specific concern arises if base64 encoding is used, because base64 is likely to have subtle timing differences that can leak information.  Similarly, a naive hex encoding implementation will leak bits when the upper nibble of a byte is zero.  Regarding other timing channels, as Adam noted, there are few implementations that treat public keys in a timing-dependent manner, but see above regarding principle.

Note that we have to be very careful about things like timing side channels in the browser.  Unintentionally leaking information to other origins, especially when it is supposed to be secret, is not good.

index.html
beverloo, line 794: I see, thanks!

In the interest of not complicating this until we have to, we could, and probably have to define the serialization value per `PushEncryptionKeyName` entry. Since both P-256 and Curve25519 have a 32 octet binary key, this can be done regardless of our decision on the curve, and will remove this ambiguity from the initial PR. The needed complexity can then be added when additional curves get supported.

martinthomson, line 794: Yes, we have to define the value on a per-key basis.

(and P-256 is actually 65 bytes.  Unless you compress, but compression isn't widely supported.)

index.html
```diff
@@ -688,6 +751,60 @@
...
+        <li>Let <var>keys</var> be an empty map.
+        </li>
+        <li>For each identifier <var>i</var> corresponding to keys in internal slots on the
+        <a><code>PushSubscription</code></a>, ordered by the name of the key:
+          <ol>
+            <li>Let <var>b</var> by the encoded value of the public key corresponding to the key
+            name <var>i</var>, using the encoding defined for the key name (see <code><a data-lt=
+            "PushSubscription.getKey">getKey</a></code>).
+            </li>
+            <li>Let <var>s</var> be the URL-safe base64 encoding of <var>b</var> as a
```
ghost, line 788: @martinthomson IIRC, you were considering constant-time hex encoding to mitigate side-channel attacks. Is that a concern with Base64 encoding?

martinthomson, line 788: Maybe.  There is no inherent reason that base64 can't be constant time, though I'll note that it commonly isn't.

martinthomson:
@mvano given that the supporting pieces have been adopted; should we merge this now?  I've updated the link to the IETF spec.

johnmellor:
Was there any resolution on how to avoid timing attacks in the endpoint and public key serialization?

martinthomson:
Endpoints, yes, I'm not aware of any way to mess up encoding of an ASCII string (which a URL is required to be).  As for the key, the confidentiality of the key is less critical (if you manage to keep the endpoint secret, then your only potential adversary is the push service).  I think that I should add some text regarding serialization here though.  Thanks for the reminder.

index.html
```diff
@@ -465,14 +479,27 @@
...
-          message</a>. A <a>push message</a> can be delivered using the <a>web push protocol</a>.
-        </p>
-        <p>
-          The <code><a>endpoint</a></code> of a <code>PushSubscription</code> is a URI that allows
-          an <a>application server</a> to request delivery of a <a>push message</a> to a <a>push
-          subscription</a>.
+          The fields included in the <code>PushSubscription</code> is all the information needed
+          for an <a>application server</a> to send a <a>push message</a>. Push services that are
+          compatible with the Push API will be expected to provide an <a>endpoint</a> that conforms
+          to the <a>web push protocol</a>. These parameters and attributes include:
```
beverloo: "Push services that...web push protocol" feels like a `<p class=note>` more than normative text. It doesn't actually say anything that the reader can do anything with.

index.html
```diff
@@ -465,14 +479,27 @@
...
         </p>
+        <ul>
+          <li>The <code><a data-lt="PushSubscription-endpoint">endpoint</a></code> of a
+          <code>PushSubscription</code> is a URI that allows an <a>application server</a> to
```
beverloo: URI -> URL?

beverloo:
LGTM % these nits. Let's get this in, we can iterate if we have to :)

martinthomson:
Nits addressed.  Assigning to @mvano

PR #131, opened 2015-03-24
Add a PushSubscriptionOptions dictionary with a userVisible option.

beverloo:
User agents may decide to display different UI when requesting push
permission from the user if the intended use of the subscription is to
only send messages which result in user visible UI, for example a
Web Notification.

This PR introduces this in the specification by accepting an optional
property bag to the subscribe() and hasPermission() methods. This bag
features a "userVisible" option that can be set to TRUE.

UAs are not required to follow this UI paradigm, but when they do, they
SHOULD enforce the requirement when receiving a push message. The exact
tools for "user visible UI" and enforcement are left undefined as
one can imagine multiple solutions for them, which we don't want to
capture in the specification.

### DIFF
--- a/index.html
+++ b/index.html
@@ -68,6 +68,18 @@
               ]
             },
           ],
+
+          localBiblio: {
+            "NOTIFICATIONS": {
+              title: "Notifications API",
+              href: "https://notifications.spec.whatwg.org/",
+              authors: [
+                "Anne van Kesteren",
+              ],
+              status: "Living Standard",
+              publisher: "WHATWG"
+            }
+          },
       };
     </script>
   </head>
@@ -313,6 +325,10 @@
         preserved beyond the current browsing session MUST be revocable.
       </p>
       <p>
+        The <a>user agent</a> MAY consider the <code><a>PushSubscriptionOptions</a></code>
+        when acquiring permission or determining the permission status.
+      </p>
+      <p>
         When a permission is revoked, all <a title="push subscription">push subscriptions</a>
         created with that permission MUST be <a>deactivated</a>.
       </p>
@@ -515,13 +531,13 @@ navigator.serviceWorker.register('serviceworker.js').then(
       </p>
       <dl title="interface PushManager" class="idl">
         <dt>
-          Promise&lt;PushSubscription&gt; subscribe ()
+          Promise&lt;PushSubscription&gt; subscribe (optional PushSubscriptionOptions options)
         </dt>
         <dt>
           Promise&lt;PushSubscription?&gt; getSubscription ()
         </dt>
         <dt>
-          Promise&lt;PushPermissionStatus&gt; hasPermission ()
+          Promise&lt;PushPermissionStatus&gt; hasPermission (optional PushSubscriptionOptions options)
         </dt>
       </dl>
       <p>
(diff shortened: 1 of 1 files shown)


beverloo:
+@martinthomson @johnmellor 

I've been careful not to mandate any UI - the option is an indication of intention. WDYT?

beverloo:
@martinthompson - All applied. Thanks for your comments!

index.html
```diff
@@ -515,13 +531,13 @@
...
         <dt>
-          Promise&lt;PushSubscription&gt; subscribe ()
+          Promise&lt;PushSubscription&gt; subscribe (optional PushSubscriptionOptions options)
         </dt>
         <dt>
           Promise&lt;PushSubscription?&gt; getSubscription ()
         </dt>
         <dt>
-          Promise&lt;PushPermissionStatus&gt; hasPermission ()
+          Promise&lt;PushPermissionStatus&gt; hasPermission (optional PushSubscriptionOptions options)
```
mounirlamouri, line 540: Out of curiosity, why is that optional?

index.html
```diff
@@ -641,6 +657,28 @@
...
+        </h2>
+        <p>
+          A <a>PushSubscriptionOptions</a> object represents additional options associated with a
+          <a>push subscription</a>. The <a>user agent</a> MAY consider these options when requesting
+          <a>express permission</a> from the user. When an option is considered, the <a>user
+          agent</a> SHOULD enforce it on incoming <a title="push message">push messages</a>.
+        </p>
+        <dl title="dictionary PushSubscriptionOptions" class="idl">
+          <dt>
+            boolean userVisible = false
```
mounirlamouri, line 672: Should that be required?

index.html
beverloo, line 540: It's perfectly fine to subscribe to push without any options (to get what we call "silent push").

index.html
beverloo, line 672: No. User agents are free to ignore these options if they have no intention of using it, hence the MAY.

Marking this as required would break forward compatibility. Imagine that Chrome uses `userVisible`, but Firefox ignores it. Instead, they introduce a `timeout` option to limit execution time of incoming push messages. Either we end up with (a) `userVisible` being required and `timeout` being optional - inconsistent, and unnecessary since Firefox doesn't use it, or (b) `userVisible` and `timeout` both being required - consistent, but breaks all current users.

beverloo:
@martinthompson - Normative text aside, are you OK with the PushSubscriptionOptions dictionary and the userVisible member? It'd be great if we can deprecate our proprietary stuff before Push gains more traction, but Chrome 43 branch point is scaringly close.

martinthomson:
Sure; I more or less expected you to merge this based on my last comments.  I don't see why you need this updated to ship Chrome changes, but maybe there is something in your process I don't understand.

On that point, I'd be much happier if you merged #129 as well, so we can move away from this being a description of a Chrome-proprietary API.

beverloo:
Your avatar is oddly relevant. :-)

Since @mvano is out for a while, we're a bit blocked on getting spec changes through. (I don't have edit rights.) That goes for #129 as well. I'll send an e-mail to @efullea..

mvano:
I'm very glad to see this one, as it really simplifies Chrome's solution, and can be used by other implementations if they want it.

Also, glad to see that this aligns with userVisible in the PushPermissionDescriptor in the Permissions API (https://w3c.github.io/permissions/#push).

mvano:
One note: we may want to expose userVisible on the subscription interface in future.

PR #148, opened 2015-05-21
Making cardinality clear

martinthomson:
We've had questions about this.  I think that this is the right answer.


mvano:
Looks good to me. Please run tidy, otherwise next time I run it after editing a different part of the document, these lines will be touched again.

PR #153, opened 2015-07-28
Removing tidy warnings with haxx

martinthomson:
Tidy warns about "</".  It doesn't warn about "<\u002f".


mvano:
Looks good to me.

PR #154, opened 2015-07-28
Protocol updates

### LINKED ISSUE #44: Is reliance on Web Push Protocol normative?
The IETF Web Push Protocol (http://datatracker.ietf.org/wg/webpush/) is developed in collaboration with the Webapps WG to complement the Push API. It is not clear to me that it should at this time be a requirement for the push service underlying the Push API to support this protocol.

My suggestion: at this time, given the early stage of the Web Push Protocol development, I think it should be referenced in a non-normative section of the Push API spec.

@beverloo @martinthomson

martinthomson:
This updates the adopted IETF work and removes some of the vestigal text from the days before we decided to have a protocol.

Closes #44.

### DIFF
--- a/index.html
+++ b/index.html
@@ -379,34 +379,30 @@
         Push Framework
       </h2>
       <p>
-        Although <a title="push service">push services</a> are expected to differ in deployment, a
-        typical deployment is expected to have the following general entities and example operation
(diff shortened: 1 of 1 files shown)


mvano:
Looks good to me. Feel free to commit when ready.

PR #155, opened 2015-07-28
Adding offline use case

martinthomson:


mvano:
Looks good to me, feel free to commit when ready.

martinthomson:
I would if I could.  Should I ask @AFBarstow to do something?

mvano:
Yes please check with Arthur. You're an editor now, so you should have the permission.

PR #156, opened 2015-08-04
Moving to using data-lt= rather than title= for a and dfn

martinthomson:
The latest respec complains about our use of title attributes for linking.  The new method is nicer.

See http://www.w3.org/respec/guide.html#definitions-and-linking

### DIFF
--- a/index.html
+++ b/index.html
@@ -97,9 +97,9 @@
   <body>
     <section id='abstract'>
       <p>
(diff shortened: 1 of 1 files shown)


mvano:
Nice. Please commit when you like. Looks like it needs a rebase.

martinthomson:
Merged manually.

PR #157, opened 2015-08-04
eventInitDict is always present when constructing an event

### LINKED ISSUE #145: The push event
Since per IDL the dictionary is always present (once you hit the API algorithm), you only have to check whether the data member is missing, not whether the eventInitDict argument is provided.

martinthomson:
Fixes #145

### DIFF
--- a/index.html
+++ b/index.html
@@ -380,30 +380,28 @@
         Push Framework
       </h2>
       <p>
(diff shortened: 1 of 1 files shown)


martinthomson:
Hmm, it looks like tidy made a mess of this changeset.

mvano:
Looks good to me, please commit when you're ready.

PR #160, opened 2015-08-21
Making PushMessageDataInit a union

### LINKED ISSUE #159: Binary messages and use of `USVString` in `PushMessageDataInit`
Section 10 states "`PushMessageData` objects have an associated bytes (a byte sequence) set on creation." Given that—and the `blob()` and `arrayBuffer()` methods—I'm curious why `PushMessageDataInit` uses `USVString` instead of, e.g., `ArrayBuffer`. Does this mean that all decrypted messages must be UTF-8 encoded strings?

martinthomson:
Fixes #159.

### DIFF
--- a/index.html
+++ b/index.html
@@ -734,7 +734,7 @@ navigator.serviceWorker.register('serviceworker.js').then(
         The <code><dfn id="widl-PushMessageData-text-USVString">text()</dfn></code> method, when
         invoked, MUST return the result of running <a>utf-8 decode</a> on <var>bytes</var>.
       </p>
-      <div class="idl" title="typedef (BufferSource or Blob or DOMString) PushMessageDataInit"></div>
+      <div class="idl" title="typedef (BufferSource or USVString) PushMessageDataInit"></div>
       <p>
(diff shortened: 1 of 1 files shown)


annevk:
This seems bad. You shouldn't be able to read a `Blob` synchronously.

martinthomson:
Hmm, you make a good point.  Should I give up on Blob?

annevk:
Yeah, and instead of DOMString you want USVString so lone surrogates get replaced.

martinthomson:
I've never understood the insistence on USVString; don't we to want to insist on inputs having no lone surrogates?

annevk:
A normal JavaScript string doesn't insist on it and DOMString goes along with that. However, whenever you have a string that needs to go over the network or become a URL they do become a problem. So at that point you'd use USVString.

martinthomson:
And I finally understand.  Updated.

I will send @heycam a PR for WebIDL, because the explanation there is woefully inadequate.

beverloo:
This seems good to me.

PR #162, opened 2015-09-03
Adding make check and circle CI support

martinthomson:
This just ensures that commits and pull requests always have tidy run on them.  That way we can avoid the manual checks we've been doing.

I used Circle CI because it has a nicer interface than Travis.  Neither were able to install or run tidy without a fair bit of extra mucking around because they run the ancient Ubuntu 12.04.

@mvano PTAL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+*~
+*.swp
+.support
+tidy-html5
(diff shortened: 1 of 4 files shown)


mvano:
Nice!

martinthomson:
@ylafon, would it be too much trouble for you to enable [Circle CI](https://circleci.com/) for this repo?

PR #171, opened 2015-11-13
Add a brief clarification about the expected X9.62 representation

beverloo:
Borrowed from webpush-wg/webpush-encryption#3


mvano:
Thanks Peter!

Issue #77, opened 2014-10-09
The term webapp seems odd


mvano:
See this discussion:
http://lists.w3.org/Archives/Public/public-webapps/2014OctDec/0077.html

I'm not sure we really need to define this at all. Is there already a reusable term for this?

annevk:
I'm not sure why this term is even used. Other API specifications don't need such a term.

martinthomson:
we discussed this and we could go with either "application" or "web application" (as in notifications API).

PR #172, opened 2015-11-16
Review use of webapp vs. Service Worker

### LINKED ISSUE #5: isRegistered should return the endpoint itself rather than a boolean

### LINKED ISSUE #15: Change references to webapp => service worker
Change references to webapp => service worker where it might be unclear that only Service Workers should use interfaces & get events etc. Other than register (because of the need for permissions UI), all interfaces and events to SW only

martinthomson:
Closes #15.

### DIFF
--- a/index.html
+++ b/index.html
@@ -359,7 +359,7 @@
         <p>
           The term <dfn>express permission</dfn> refers to an act by the user, e.g. via user
           interface or host device platform features, via which the user approves the use of the
-          push API by the <a>webapp</a>.
+          Push API by the <a>webapp</a>.
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -308,8 +308,8 @@
         </p>
         <p>
           A <a>push message</a> is delivered to the <a>active worker</a> associated with the
-          <a>push subscription</a> to which the message was submitted. If the worker or its
-          <a>webapp</a> is not currently running, the worker is started to enable delivery.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -358,8 +358,8 @@
         </h2>
         <p>
           The term <dfn>express permission</dfn> refers to an act by the user, e.g. via user
-          interface or host device platform features, via which the user approves the permission of
-          a <a>webapp</a> to access the Push API.
+          interface or host device platform features, via which the user approves the use of the
+          push API by the <a>webapp</a>.
```
mvano: nit: Push API

mvano:
I'm not sure how this relates to #5, but #15 and #77 both seem relevant. This seems like a step in the right direction.

martinthomson:
My bad, I meant #15.  Nit fixed too.

PR #174, opened 2015-11-16
Add authentication secret

martinthomson:
This adds the API pieces in support of webpush-wg/webpush-encryption#4.

In short, `getKey("auth")` returns a new value that the application server is required to mix into the key derivation for encryption.


mvano:
Looks good to me if agl is happy with https://github.com/webpush-wg/webpush-encryption/pull/4

martinthomson:
Updated, new basis at webpush-wg/webpush-encryption#4 (edit, same location as before, different base).

PR #175, opened 2015-11-23
Only expose PushMessageData in Service Workers

beverloo:


johnmellor:
Looks good to me - these objects are only provided in Service Worker push events.

PR #180, opened 2016-01-20
Voluntary server identification, WebCrypto version

martinthomson:
This uses WebCrypto constructs for describing the key.  I'm starting to think that we should just use a raw key (BufferSource) here and just cite X9.62.

@beverloo, any preferences?


martinthomson:
#182 is better.

PR #191, opened 2016-05-04
Note that the subscription shares fate with the registration

### LINKED ISSUE #190: Clarify when to unsubscribe if the service worker is unregistered
As @wanderview points out in https://bugzilla.mozilla.org/show_bug.cgi?id=1185716#c20, it's possible for a client to be using a service worker registration when it's removed. In that case, the registration is flagged, and can be restored or removed later.

It's unclear what to do with the push subscription in this case. Should it be removed unconditionally, even when the service worker still has clients? Or should it only be removed once all clients have gone away?

martinthomson:
For @wanderview

Closes #190.

### DIFF
--- a/index.html
+++ b/index.html
@@ -261,7 +261,9 @@
         <code><a href=
         "http://www.w3.org/TR/service-workers/#extendable-event-init-dictionary"><dfn>ExtendableEventInit</dfn></a></code>,
         <a href="http://www.w3.org/TR/service-workers/#dfn-extend-lifetime-promises"><dfn>extend
-        lifetime promises</dfn></a> and the <a href=
+        lifetime promises</dfn></a>, the <a href=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -657,6 +657,12 @@
         subscription</a>.
         </li>
       </ol>
+      <p class="note">
+        A <code><a>PushSubscription</a></code> is associated with a <a>service worker
```
mvano: I think most of this is captured already in http://w3c.github.io/push-api/#push-subscription in normative form, which I prefer over notes, which are non-normative. Maybe just clarify that section by specifically referring to the "clear registration" algorithm?

martinthomson:
Updated as @mvano suggests.

mvano:
Sweet, thanks!

PR #192, opened 2016-05-12
Use NotAllowedError instead of defining a new PermissionDeniedError. …

### LINKED ISSUE #107: Refer to reusable definition of PermissionDeniedError once that becomes available
There is an open bug for defining PermissionDeniedError [1] and once it is defined the Push API should just refer to that.

[1] https://www.w3.org/Bugs/Public/show_bug.cgi?id=23033

martinthomson:
… Closes #107.


mvano:
Neat, thanks!

PR #195, opened 2016-06-09
Acknowledge messages that fail to decrypt

martinthomson:
There is no good way to recover from decryption failure, so the
browser might as well acknowledge them.  That will prevent the
push service from sending them again, which is good.  The main
drawback is that it looks to the application server like they
were successful.

This came up in discussions at the IETF.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1065,10 +1065,21 @@ navigator.serviceWorker.register('serviceworker.js').then(
               associated with <var>subscription</var> and the process described in
               [[!WEBPUSH-ENCRYPTION]]. This produces the plain text of the message.
               </li>
-              <li>If the <a>push message</a> could not be decrypted for any reason, discard the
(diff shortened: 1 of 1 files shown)


mvano:
Would it be possible to acknowledge receipt AND indicate it could not be decrypted? Simply acknowledging receipt would show up in logs as a successful push, and it's really not.

martinthomson:
Sadly, no.  There have been proposals for sending back information about errors, but that actually creates a reverse channel from the user agent to the application server and we don't know how to secure that.  We concluded that there needed to be stronger justification for the use case.  This was discussed at some length on webpush-wg/webpush-protocol#42 and at in-person meetings.  I think that we might eventually have something here, but we haven't worked out a good solution yet.

mvano:
I see. Protecting the device from unlimited retries seems like a necessary protection, so I'm ok with this change.

PR #201, opened 2016-07-14
Options are...optional.

### LINKED ISSUE #200: Allow UA to not implement all PushSubscriptionOptions
These need to be truly optional.  If a UA doesn't implement X, then it shouldn't be obliged to remember values that were set by applications.

martinthomson:
Closes #200.

@beverloo, I decided to go with MUST NOT.  That keeps things copacetic I think.


beverloo:
LGTM :+1:

PR #217, opened 2016-11-03
Update reference to DSS spec

### LINKED ISSUE #210: update reference to FIPS-186-4
[FIPS-186-3] on the reference list has been superseded by FIPS-186-4 published in July 2013.  [WEBPUSH-ENCRYPTION] references FIPS-186-4 already.  It'd be nice if the spec could be updated to keep consistency with [WEBPUSH-ENCRYPTION].  BTW, there is no technical difference in terms of how P-256 should work.

martinthomson:
Closes #210


beverloo:
specref.org says yes -- thanks!

PR #218, opened 2016-11-03
Rewrite introduction to clarify applicability and use cases

### LINKED ISSUE #179: Add applicability statement
Paul Banks notes that we aren't very clear about what Push is good for.  Something like this is probably a good idea:

> The Push API is intended for infrequent messages.  If you have a page
> open to site, it might still be suitable to rely on push for messages
> that are infrequent or unpredictable.  However, if you are actively
> communicating with your site, it is best to use more direct means of
> sending messages, such as HTTPS.

martinthomson:
There were a lot of cases in the intro that hinted at use cases that push really isn't suited to.  In fact, many of the use cases were not being addressed by the API at all.  Thus, I cut things down a lot and concentrated on two:

* the webapp isn't open
* the browser is offline

I then go on to explain how messages can be sent at any time, but that push messaging is relatively inefficient (and why).

This probably needs a careful review.

Closes #179.

### DIFF
--- a/index.html
+++ b/index.html
@@ -152,13 +152,12 @@
       </h2>
       <p>
         The Push API allows a <a>webapp</a> to communicate with a <a>user agent</a> asynchronously.
-        This allows a <a>webapp</a> to provide the <a>user agent</a> with time-sensitive
-        information whenever that information becomes known, rather than waiting for a user to open
-        the <a>webapp</a>.
+        This allows an <a>application server</a> to provide the <a>user agent</a> with
+        time-sensitive information whenever that information becomes known, rather than waiting for
+        a user to open the <a>webapp</a>.
       </p>
       <p>
-        As defined here, <a>push services</a> support delivery of <a>application server</a>
-        messages at any time.
+        As defined here, <a>push services</a> support delivery of <a>push messages</a> at any time.
       </p>
       <p>
         In particular, a <a>push message</a> will be delivered to the <a>webapp</a> even if that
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -151,47 +151,47 @@
         Introduction
       </h2>
       <p>
+        The Push API allows a <a>webapp</a> to communicate with a <a>user agent</a> asynchronously.
+        This allows a <a>webapp</a> to provide the <a>user agent</a> with time-sensitive
+        information whenever that information becomes known, rather than waiting for a user to open
+        the <a>webapp</a>.
```
beverloo: We define _webapp_ as the code executed by the UA, where _application server_ is the server-side component of a _webapp_.

As such, it's the _application server_ that is enabled to communicate with the _webapp_, and thereby the _user agent_, asynchronously. It seems to me like the first two occurrences of _webapp_ have to be _application server_, and the occurrences of _user agent_ should be _webapp_?

index.html
```diff
@@ -151,47 +151,47 @@
...
+      </p>
+      <p>
         As defined here, <a>push services</a> support delivery of <a>application server</a>
-        messages in the following contexts and related use cases:
+        messages at any time.
```
beverloo: > ... support delivery of <a>push messages</a> at any time.

index.html
```diff
@@ -151,47 +151,47 @@
...
+        learns of changes that occur while a user is offline and ensures that the <a>user agent</a>
+        can be provided with relevant information in a timely fashion. For example, a user that is
+        away from network access might receive a new message; the message is stored by the <a>push
+        service</a> until the <a>user agent</a> becomes reachable and the message can be delivered.
+      </p>
+      <p>
+        The Push API will also ensure reliable delivery of push messages while a <a>user agent</a>
+        is actively using a <a>webapp</a>, for instance if a user is actively using the
+        <a>webapp</a> or the <a>webapp</a> has an active worker, frame, or background window.
+        However, as mentioned, these are not the primary use case for the API.
```
beverloo: nit: s/use case/use cases/

Is it worth emphasizing that while they're not the primary use cases, it's still something developers should consider when the frequency of messages is low enough? We shouldn't advocate the WebRTC call application to maintain a connection of their own when all it does is waiting for incoming calls.

index.html
```diff
@@ -151,47 +151,47 @@
...
+      </p>
+      <p>
+        Push messaging is best suited to occasions where there is not already an active
+        communications channel established between <a>user agent</a> and <a>webapp</a>. Sending a
+        <a>push message</a> requires considerably more resources when compared with more direct
+        methods of communication such as <a href="https://fetch.spec.whatwg.org/">fetch()</a> or
+        <a href="https://html.spec.whatwg.org/multipage/comms.html#network">websockets</a>. <a>Push
+        message</a> usually have higher latency than direct communications and they can also be
```
beverloo: nit: s/Push message/Push messages/

index.html
```diff
@@ -151,47 +151,47 @@
...
+        message</a> might be used to inform the user of an incoming WebRTC call.
+      </p>
+      <p>
+        A <a>push message</a> can also be sent when the <a>user agent</a> is temporarily offline.
+        In support of this, the <a>push service</a> stores messages for the <a>user agent</a> until
+        the <a>user agent</a> becomes available. This supports use cases where a <a>webapp</a>
+        learns of changes that occur while a user is offline and ensures that the <a>user agent</a>
+        can be provided with relevant information in a timely fashion. For example, a user that is
+        away from network access might receive a new message; the message is stored by the <a>push
+        service</a> until the <a>user agent</a> becomes reachable and the message can be delivered.
```
beverloo: "For example, a user that...can be delivered." just repeats the previous sentence. I think we can drop it.

index.html
```diff
@@ -151,47 +151,47 @@
...
+        However, as mentioned, these are not the primary use case for the API.
+      </p>
+      <p>
+        Push messaging is best suited to occasions where there is not already an active
+        communications channel established between <a>user agent</a> and <a>webapp</a>. Sending a
+        <a>push message</a> requires considerably more resources when compared with more direct
+        methods of communication such as <a href="https://fetch.spec.whatwg.org/">fetch()</a> or
+        <a href="https://html.spec.whatwg.org/multipage/comms.html#network">websockets</a>. <a>Push
+        message</a> usually have higher latency than direct communications and they can also be
+        subject to restrictions on use. Most push services limit the size and quantity of <a>push
```
beverloo: nit: s/push services/<a>push services</a>/

beverloo (commented):
This looks great, thanks!

PR #220, opened 2016-11-08
Add Peter Beverloo as an editor, retire Michael

### LINKED ISSUE #219: Update editors
@mvano might not be active
@beverloo will be more active

beverloo:
Fixes #219. Thanks for all your hard work, Michael!

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,16 +21,23 @@
 
           editors: [
             {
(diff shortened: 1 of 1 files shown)


mvano:
Thanks Peter for taking this on!

PR #224, opened 2016-11-18
Clarify encryption key lifetime and intention

### LINKED ISSUE #212: Some things about the PushEncryptionKeys p256dh and auth are not clear enough
In the working draft 15 is written: 

> Keys named p256dh and auth must be supported.

Does that mean that those keys must exist and contain a value, is it enough that they exist and contain null or does that mean (in context of what comes before that sentence) that if getKey() is supported, only then you must be able to get auth and p256dh?

Secondly, what is not clear to me is, whether it is possible that auth and p256dh can change after the subscription:
Case 1: If the values were null, is it possible that a value gets added without a new subscription - For example: a browser wasn't able to generate these values and was now updated to a version which can
Case 2: A browser had values, is it possible that these get deleted or changed without a new subscription?
So basically the question is can I assume that in any case the subscription settings stay exactly the same until the endpoint is invalidated?

In the editor draft from 18 August 2016 is written: 

> Once set, options for a push subscription cannot change.

But that is in a part which focuses on the  PushSubscriptionOptions userVisibleOnly and applicationServerKey. And if that also contains p256 and auth, then the question is again: does "set" mean filled with a value or does null qualify too, that the "value" cannot be changed?

Thanks

beverloo:
Fixes #212

### DIFF
--- a/index.html
+++ b/index.html
@@ -350,6 +350,12 @@
           most one <a>push subscription</a>.
         </p>
         <p>
+          A <a>push subscription</a> has internal slots for a P-256 <a>ECDH</a> key pair and an
+          authentication secret in accordance with [[!WEBPUSH-ENCRYPTION]]. These slots MUST be
+          populated when creating the <a>push subscription</a>, and MUST remain constant for its
+          lifetime.
+        </p>
+        <p>
           When a <a>push subscription</a> is <dfn data-lt=
           "deactivated|deactivate">deactivated</dfn>, both the <a>user agent</a> and the <a>push
           service</a> MUST delete any stored copies of its details. Subsequent <a>push messages</a>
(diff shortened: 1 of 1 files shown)


beverloo:
FYI @PascalReintjes

I'm fiddling around with introducing a concept of "invalidating" push subscriptions in preparation of the bigger `pushsubscriptionchange` PR. That's definitely relevant since we'll want to detail what happens when the UA has to change the encryption keys, but I don't think that has to block the clarifications.

martinthomson:
Thanks Peter, I agree that the invalidation text really does need to go with the event refactor work.

PascalReintjens:
Thanks @beverloo, that makes it perfectly clear :)

Issue #209, opened 2016-07-25
applicationServerKey is never used in subscribe()


jyasskin:
[PushManager.subscribe()](https://w3c.github.io/push-api/#widl-PushManager-subscribe-Promise-PushSubscription--PushSubscriptionOptionsInit-options) checks `applicationServerKey` for validity, but it never actually uses the value.

martinthomson:
That is correct, the browser doesn't actually use this value.  It should send this to the push service though.  We should fix that.

PR #227, opened 2016-11-25
Allow passing a base64url-encoded value to `applicationServerKey`

### LINKED ISSUE #226: Ergonomics of VAPID key
The [applicationServerKey](https://w3c.github.io/push-api/#widl-PushSubscriptionOptions-applicationServerKey) is currently specified as a `BufferSource`.  We could be a little more flexible in terms of how we allow applications to set this value.  

Options:

1. Keep it as is
2. Add a base64url option (make the type a union of DOMString and BufferSource)
3. Something else

@costinm suggested that the browser could acquire this key automatically from a `.well-known` location on the origin.  I think that adds a level of brittleness to the API that isn't warranted.  A variation on this is to tell the push service about the origin of the page and have it do the lookup.  That's a privacy regression in addition to being brittle.

Option 2 seems fairly easy.  It might mean that the browser can avoid having to encode into a string for sending to the server.  In practice, it isn't any improvement because we recommend that user agents validate the key (which Firefox does).

cc @jrconlin @costinm

beverloo:
We give developers a base64url-encoded value when they JSON serialize
the PushSubscription object, and also use the base64url-encoded
representation of public keys when using VAPID for application server to
push service communication. This PR aligns the JavaScript API with that.

Fixes #226

### DIFF
--- a/index.html
+++ b/index.html
@@ -248,6 +248,8 @@
         <code><a href=
         "http://www.w3.org/TR/dom/#invalidaccesserror"><dfn>InvalidAccessError</dfn></a></code>,
         <code><a href=
+        "http://www.w3.org/TR/dom/#invalidcharactererror"><dfn>InvalidCharacterError</dfn></a></code>,
+        <code><a href=
         "http://www.w3.org/TR/dom/#securityerror"><dfn>SecurityError</dfn></a></code>,
         <code><a href="http://www.w3.org/TR/dom/#networkerror"><dfn>NetworkError</dfn></a></code>,
         <a href="http://www.w3.org/TR/dom/#concept-event-listener"><dfn>event listener</dfn></a>,
@@ -297,6 +299,8 @@
         "http://heycam.github.io/webidl/#idl-ArrayBuffer"><dfn>ArrayBuffer</dfn></a></code>,
         <code><a href=
         "http://heycam.github.io/webidl/#common-BufferSource"><dfn>BufferSource</dfn></a></code>,
+        <code><a href=
+        "https://heycam.github.io/webidl/#idl-DOMString"><dfn>DOMString</dfn></a></code>,
         <code><a href=
         "http://heycam.github.io/webidl/#notallowederror"><dfn>NotAllowedError</dfn></a></code>,
         and <code><a href=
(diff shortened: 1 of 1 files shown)


beverloo:
I considered addressing #209 as part of this, but realized that VAPID's restrictions depend on `Crypto-Key` which is now gone. Do you have ideas on what the succession of that should be?

martinthomson:
No, it's the main thing preventing me from publishing a new version.  Ideas sought.

index.html
```diff
@@ -601,10 +603,22 @@
...
-        an <code><a>InvalidAccessError</a></code> and terminate these steps.
+        <code><a>applicationServerKey</a></code> attribute, run the following substeps:
+          <ol>
+            <li>Let <var>applicationServerKey</var> be the sequence of octets in
+            <code><a>applicationServerKey</a></code> when provided as a <code><a>BufferSource</a></code>,
+            or the sequence of octets that results from decoding
+            <code><a>applicationServerKey</a></code> using the URL-safe base64 encoding [[!RFC4648]]
+            when provided as a <code><a>DOMString</a></code>. If decoding fails, reject
+            <var>promise</var> with a <code><a>DOMException</a></code> whose name is
+            "<code><a>InvalidCharacterError</a></code>" and terminate these steps.
```
martinthomson, line 610: What do we do about padding (the trailing '===')?

index.html
```diff
@@ -601,10 +603,22 @@
...
-        <code><a>applicationServerKey</a></code> attribute, check that the value is valid (i.e.,
-        ensure that it describes a valid point on the P-256 curve). If the
-        <code><a>applicationServerKey</a></code> value is invalid, reject <var>promise</var> with
-        an <code><a>InvalidAccessError</a></code> and terminate these steps.
+        <code><a>applicationServerKey</a></code> attribute, run the following substeps:
+          <ol>
+            <li>Let <var>applicationServerKey</var> be the sequence of octets in
+            <code><a>applicationServerKey</a></code> when provided as a <code><a>BufferSource</a></code>,
+            or the sequence of octets that results from decoding
+            <code><a>applicationServerKey</a></code> using the URL-safe base64 encoding [[!RFC4648]]
```
martinthomson: You can use RFC 7515 and base64url, which is a more precise encoding (and one that addresses my next comment).

beverloo: Done (also below).

PR #231, opened 2016-12-07
chore: use new school WebIDL

marcoscaceres:


marcoscaceres:
@beverloo, this is 95% done, but I might need you to do the last bits. In particular, you need to split dictionary and interfaces into their own sections (and then just follow what I did). The reason is that there are things that are defined in both interfaces and dictionaries (e.g., "applicationserverkey"). You can explicitly differentiate and reference between the two by either:

```HTML
The <a>WhateverDictionary.applicationserverkey</a> does x. 
```

Or:

```HTML
The <a data-lt="WhateverDictionary.applicationserverkey">applicationserverkey</a> does x. 
```

See also how I've added `<section data-dfn-for="Thing">` and how `<dfn>`s are defined. It should be quite straight forward.

index.html
```diff
@@ -205,7 +205,7 @@
     <section id="conformance">
       <p>
         This specification defines conformance criteria that apply to a single product: the
-        <dfn data-lt="user agent|user agents">user agent</dfn> that implements the interfaces that
+        <dfn data-lt="agents|user agents">user agent</dfn> that implements the interfaces that
```
marcoscaceres, line 208: note for data-lt's, you don't need to repeat the defined term - so I removed those.

Issue #193, opened 2016-05-24
Pass removed subscriptions as a property of the `pushsubscriptionchange` event


ghost:
@collimarco points out in http://blog.pushpad.xyz/2016/05/the-push-api-and-its-wild-unsubscription-mechanism/ that we don't currently pass the removed subscription when we fire `pushsubscriptionchange`. At that point, `self.registration.pushManager.getSubscription()` will resolve to `null`, because the old subscription has been removed.

An app can work around this by storing the subscription endpoint in IDB, then retrieving it when `pushsubscriptionchange` is fired. Alternatively, we can pass the old subscription along, maybe as a property on the event:

``` webidl
interface PushSubscriptionEvent : ExtendableEvent {
  readonly attribute PushSubscription subscription;
}
```

martinthomson:
That might be confused for the new subscription.  Maybe call it `oldSubscription`.  Also note that we might not actually return `null` in response to `getSubscription()` if we are pre-emptively removing the subscription (the protocol doesn't support a signal of this sort, but we know that no one is currently using the protocol on this leg just yet).

collimarco:
+1 (obviously)

I would like to remark that this issue should apply to the following cases:
- the user blocks the notifications (a single pushsubscriptionchange is triggered passing the old subscription so that it can be immediately removed from the app server)
- the push service invalidates the endpoint (a single pushsubscriptionchange is triggered passing the old subscription and a new endpoint can be obtained with `getSubscription()`, so that the old endpoint can be replaced with the new one on the server preserving attached data)

ghost:
> Also note that we might not actually return `null` in response to `getSubscription()` if we are pre-emptively removing the subscription

Yeah, I think @johnmellor suggested this in https://github.com/w3c/push-api/issues/132#issuecomment-189308012. At that point, a "change" could mean a few things:
1. The subscription is still active, but going away soon. The app should resubscribe soon, and the old subscription will be deactivated once the new subscription is used.
2. The user revoked permission, or the subscription exceeded the background message quota (in Firefox). The app can't resubscribe, but can still notify its server to clean up the old subscription. It's OK if the app ignores this event.
3. The server removed the subscription, the user restored a previously-revoked permission, or reset the quota by visiting the origin again. The app should definitely listen for this event so that it can resubscribe.

[...]

Issue #132, opened 2015-03-25
Describe process for subscription updates


martinthomson:
Any time a subscription needs to change, we signal that with the `subscriptionchange` event.  Some of these cases won't be as a result of a break and it would be good to describe how a subscription is seamlessly updated.

I think that the process is simple:
1. Determine that a subscription update is needed for any reason.  This might be expiration of the subscription, or a desire to roll out the old encryption keys (see #55).
2. Create a new subscription with the new parameters, while leaving the old subscription in place.
3. Fire the event as an `ExtendableEvent`.
4. Await the `waitUntil` promise if that was set by the application.
5. Clean up the state from the old subscription.

Note that the new subscription can share properties with the old subscription; but this process is necessary for any change where the application might have to share this information with the application server. Keeping the old information around until the application has confirmed that this has happened means that there won't need to be messages lost.  We might also note that sometimes it can take some time for the update to happen.

johnmellor:
Currently the `pushsubscriptionchange` event is not specced to automatically create a new subscription, instead the spec just says:

> A Service Worker SHOULD attempt to resubscribe while handling this event, in order to continue receiving push messages.

But it's unclear how an author can "resubscribe". Should they unsubscribe then subscribe again, or just call subscribe again?

A possible process for reliably updating the subscription, even on flaky networks, would be:
1. Push service informs UA that the subscription will soon be invalidated.
2. UA wakes up SW and fires `pushsubscriptionchange`.
3. Calls to `PushManager.getSubscription` will still resolve with the old subscription for now.
4. The next time `PushManager.subscribe` is called, the subscription stored by the UA is replaced by the updated subscription, and the `subscribe` promise is resolved with this new subscription. Subsequent calls to `getSubscription` will resolve with the new subscription.
5. The website should now transmit the updated subscription to its server, retrying across restarts until it succeeds.
6. Both the old and new subscriptions are simultaneously stored by the push service, and messages can be sent using either. But once the new subscription is used for the first time, the old subscription will be deleted from the push service (after a short timeout, to allow for propagation delay in the application server(s)).
7. The old subscription will eventually expire even if no new subscription is made, or the new subscription is never used, but only after a generous grace period (weeks? months?).

Concretely, this would require changes like the following to the Push API:

8. PushManager.subscribe method steps:
[...]

Issue #120, opened 2015-03-04
pushsubscriptionchange - Initial feedback


gauntface:
After talking with @johnmellor and @jeffposnick on different occasions a few concerns have come up that may be worth discussing the appropriate approach from a developer approach or see if the API simply is the best approach.
- It feels like the pushsubscriptionchange event results in a somewhat poor experience for the user if there is any kind of error. If the subscription changes, the developer re-subscribes (assuming the permission state has changed), tries to send this to their server, that fails for whatever reason, now the developer has no way out other than to wait for the user to visit their page again in the hope that they can then send the new subscription.
- @jeffposnick had some concerns with respect to performing an update of the subscriptionId and endpoint when OAuth is needed to authenticate the user.

I understand that in Chromes current implementation this isn't a problem since GCM doesn't have a need to change the subscriptionId (at least to the best of my knowledge), but for future implementors, this feels like the API will result in a number of bad patterns by developers or poor experiences for users (i.e. randomly opening windows to update the subscription details OR the user simply not getting the push messages because the subscription could be updated on the server).

Is it worth considering moving this to something that can only occur when the user is on the web page?

martinthomson:
I think that the primary concern here is the window of time between when a subscription expires or breaks and the application server is updated with new details.  A lot happens in that time potentially.

On the positive side, a lot of that can be masked by the browser, at least for expiration.  The browser can continue to maintain the old subscription until the event it delivers to the SW has been resolved (that is, if `e.waitUntil()` is used properly).  In the case of a break, the outage window is unavoidable.

There is no need to open windows, though. That would simply be bad design.  Just `fetch()` the new details to the server (ugh, that's an awkward turn of phrase).

I think that perhaps the above should be made part of the algorithm for delivering `pushsubscriptionchange` events.

p.s., I believe that GCM does change identifiers in rare occasions (i.e., if a node crashes, but I guess that would happen very rarely).

gauntface:
I think you have pointed out the main detail I've failed to spot, if the
event.waitUntil fails to resolve then the old subscriptionId would remain
unable.

beverloo:
PR #234 defined this in a new `Subscription Refreshes` section. We suggest using background sync for getting the updated details to your server (which has yet to be changed), and don't _require_ the old subscription to be immediately invalidated exactly to allow for a brief window during which the change can propagate.

Issue #61, opened 2014-10-02
push-register (pushregistrationlost / pushsubscriptionchange) behavior when clearing browser cache


beverloo:
The specification does not define whether the `push-register` (or `pushregistrationlost`) event should be invoked when a user clears their browser cache/cookies.

While this is part of the larger question of Push + browser cache problem, it would be good to nail down for the `push-register` event and in the privacy section what the expected behavior is.

Straw-man: We must not invoke `push-register` (or `pushregistrationlost`) when the user clears their browser cache. The UA could provide for an interface themselves to clarify to the user that expected functionality might be lost.

johnmellor:
If clearing the cache clears Service Workers, then there wouldn't be anywhere to deliver the event to anyway.

martinthomson:
Any identifiers that are generated by the API need to be cleared when cookies are cleared.  Clearing cookies is usually a trigger for removing tracking mechanisms, and registration identifiers are definitely tracking mechanisms.

beverloo:
We have this covered in the `Security and privacy considerations` section.

> When a service worker registration is unregistered, any associated push subscription must be deactivated.

Clearing data deletes the Service Worker registration, which deactivates the push subscription, which (eventually) tells the push service, that then can tell the application server.

delapuente:
How exactly does the push service tell the application server that the subscription has been deactivated?

ghost:
@delapuente The app server will get a 404 or a 410 when it tries to send to the subscription.

delapuente:
Thank you.

PR #234, opened 2016-12-20
Properly define the pushsubscriptionchange event

### LINKED ISSUE #228: Properly defining `pushsubscriptionchange`
Hi all —

The `pushsubscriptionchange` event ([ref](https://w3c.github.io/push-api/)) is rather poorly defined right now. There are several issues requesting changes and clarifications. I'd like to propose the following semantics for the event.

There are three scenarios in which the browser may want to fire the event.

1. **Subscription invalidation**. This happens when any property of the subscription changes: a new endpoint (expiration and/or refresh) or new encryption keys. The UA should resubscribe on behalf of the developer _before_ firing the event. The old subscription should continue to work until either the new subscription has been used, or a certain period of time passes, e.g. 90 days.
2. **Subscription revocation**. This happens when the UA has dropped the subscription: permission got revoked (#116) or suspended because the website ran out of quota. The website cannot resubscribe in this case.
3. **Subscription lost**. This happens in situations where the UA loses the subscription: database corruption on the PS, or partially on the UA. The UA should attempt to resubscribe on behalf of the developer _before_ firing the event.

Clearing data (#61) is explicitly excluded from scenario (2). We should not inform websites when that happens, even though the application server can still detect this as the push service should reject new messages if/when the unsubscription has propagated. I consider Firefox' invocation when permission is regranted to be equivalent to scenario (3).

I think that we can support these scenarios with the following interface (#132).

The `oldSubscription` attribute (#193) would be NULL in the _subscription lost_ scenario when it's known that there was a subscription, but no complete `PushSubscription` object could be created.

The `newSubscription` attribute would be NULL in the _subscription revocation_ case, as well as in the _subscription lost_ case when resubscribing fails due to insufficient knowledge (i.e. `applicationServerKey`) or networking errors.

```webidl
interface PushSubscriptionChangeEvent : ExtendableEvent {
    readonly attribute PushSubscription? newSubscription;
    readonly attribute PushSubscription? oldSubscription;
};
```

We'll advocate use of [Background Sync](https://wicg.github.io/BackgroundSync/spec/) (#120) for transferring the updated subscription information to the server. This enables us to deal with failed fetches without effectively duplicating that entire specification. Note that Background Sync currently does not support this use-case due to restrictions: it requires a top-level frame in order to register. We'll propose changing the spec.

On a tangent, two changes I'd like to make in Chrome to help developers are (a) issuing a console warning when there's a listener for the `push` event, but not the `pushsubscriptionchange` event, and (b) adding a button to DevTools for manually triggering a `pushsubscriptionchange` event.

Before I go off and update the spec, I'd love your feedback on this :).

/cc @martinthomson @kitcambridge @gauntface @ShijunS

beverloo:
This is a first attempt at #228 and related issues. An [online preview](https://rawgit.com/beverloo/push-api/change-event/index.html) is available.

I added a few inline issues because iteration will be necessary - more things will have to be split out in algorithms. I think that coincidentally improves clarity, so I plan to model the `push` event after `pushsubscriptionchange`, but will do that in a separate PR.

WDYT?

Closes #228, #193, #132, #120, #61.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1158,14 +1158,15 @@ navigator.serviceWorker.register('serviceworker.js').then(
           refreshed, revoked or lost.
         </p>
         <p>
-          To <dfn>fire the pushsubscriptionchange event</dfn> given <var>registration</var>,
-          <var>newSubscription</var> and <var>oldSubscription</var>, the <a>user agent</a> must run
-          the following steps:
+          To <dfn>fire the pushsubscriptionchange event</dfn> given a
+          <a>service worker registration</a> of <var>registration</var>, <var>newSubscription</var>
+          and <var>oldSubscription</var>, the <a>user agent</a> must run the following steps:
         </p>
         <ol>
           <li>Create a <a>trusted event</a>, <var>event</var>, that uses the
-          <a>PushSubscriptionChangeEvent</a> interface, with the event type <code>pushsubscriptionchange</code>,
-          which does not bubble, is not cancelable, and has no default action.
+          <a>PushSubscriptionChangeEvent</a> interface, with the event type
+          <code>pushsubscriptionchange</code>, which does not bubble, is not cancelable, and has no
+          default action.
           </li>
           <li>Set the <code>newSubscription</code> attribute of <var>event</var> to
           <var>newSubscription</var>.
@@ -1173,9 +1174,8 @@ navigator.serviceWorker.register('serviceworker.js').then(
           <li>Set the <code>oldSubscription</code> attribute of <var>event</var> to
           <var>oldSubscription</var>.
           </li>
-          <li>Invoke the <a>Handle Functional Event</a> algorithm with <var>event</var>, a
-          <a>service worker registration</a> of <var>registration</var> and <var>callbackSteps</var>
-          set to the following steps:
+          <li>Invoke the <a>Handle Functional Event</a> algorithm with <var>event</var>,
+          <var>registration</var> and <var>callbackSteps</var> set to the following steps:
             <ol>
               <li>Set <var>global</var> to the global object associated with the
               <var>registration</var>.
@@ -1188,7 +1188,8 @@ navigator.serviceWorker.register('serviceworker.js').then(
         <p class="note">
           Consider using a more reliable synchronization mechanism such as [[WEB-BACKGROUND-SYNC]]
           when sending the details of the new <a>push subscription</a> to your
-          <a>application server</a>. The user might be subject to unreliable network conditions.
+          <a>application server</a>. The user might be subject to unreliable network conditions that
+          could cause a fetch to fail.
         </p>
 
         <section data-dfn-for="PushSubscriptionChangeEvent">
--- a/index.html
+++ b/index.html
@@ -473,8 +473,13 @@
         acquiring permission or determining the permission status.
       </p>
       <p>
-        When a permission is revoked, all <a>push subscriptions</a> created with that permission
-        MUST be <a>deactivated</a>.
+        When a permission is revoked, the <a>user agent</a> MAY
+        <a>fire the pushsubscriptionchange event</a> for subscriptions created with that permission,
+        with the <a>servce worker registration</a> associated with the <a>push subscription</a> as
+        <var>registration</var>, a <a>PushSubscription</a> instance representing the
+        <a>push subscription</a> as <var>oldSubscription</var>, and <code>null</code> as
+        <var>newSubscription</var>. The <a>user agent</a> MUST, in parallel, <a>deactivate</a>
+        the affected subscriptions.
       </p>
       <p>
         When a <a>service worker registration</a> is unregistered, any associated <a>push
--- a/index.html
+++ b/index.html
@@ -1150,59 +1150,86 @@ navigator.serviceWorker.register('serviceworker.js').then(
       </section>
       <section>
         <h2>
-          The "<dfn>pushsubscriptionchange</dfn>" event
+          The <dfn>pushsubscriptionchange</dfn> event
         </h2>
         <p>
-          The <a>pushsubscriptionchange</a> event indicates that a <a>push subscription</a>
-          has been invalidated, or will soon be invalidated. For example, the <a>push service</a>
-          MAY set an expiration time. A <a>Service Worker</a> SHOULD attempt to resubscribe while
-          handling this event, in order to continue receiving <a>push messages</a>.
+          The <a>pushsubscriptionchange</a> event indicates a change in a <a>push subscription</a>
+          that was triggered outside of the application's control, for example because it has been
+          refreshed, revoked or lost.
         </p>
         <p>
-          When new <a>push subscription</a> information becomes available, the <a>user agent</a>
-          MUST run the following steps:
+          To <dfn>fire the pushsubscriptionchange event</dfn> given <var>registration</var>,
+          <var>newSubscription</var> and <var>oldSubscription</var>, the <a>user agent</a> must run
+          the following steps:
         </p>
         <ol>
-          <li>Let <var>registration</var> be the <a>service worker registration</a> corresponding
-          to the push message.
+          <li>Create a <a>trusted event</a>, <var>event</var>, that uses the
+          <a>PushSubscriptionChangeEvent</a> interface, with the event type <code>pushsubscriptionchange</code>,
+          which does not bubble, is not cancelable, and has no default action.
           </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -359,30 +359,80 @@
...
+          it MUST <a>fire the pushsubscriptionchange event</a> with the
+          <a>service worker registration</a> associated with the <a>push subscription</a> as
+          <var>registration</var>, a <a>PushSubscription</a> instance representing the
+          <a>push subscription</a> having the old keys as <var>oldSubscription</var> and a
+          <a>PushSubscription</a> instance representing the <a>push subscription</a> having the new
+          keys as <var>newSubscription</var>.
+        </p>
+        <section>
+          <h2>
+            Subscription refreshes
```
martinthomson: nit: 
title case for titles (throughout)

index.html
```diff
@@ -359,30 +359,80 @@
...
+          </p>
+          <p>
+            When this happens, the <a>user agent</a> MUST create a new <a>push subscription</a>
+            with the <a>push service</a> on behalf of the application, using the
+            <a>PushSubscriptionOptions</a> that were provided for creating the current
+            <a>push subscription</a>. The new <a>push subscription</a> MUST have a key pair that's
+            different from the original subscription.
+          </p>
+          <p>
+            When successful, <a>user agent</a> then MUST <a>fire the pushsubscriptionchange event</a>
```
martinthomson, line 398: What about when the refresh is not successful?  I think that we could recommend/suggest that the user agent either retain the old subscription, though it may treat the subscription as broken at its discretion.

index.html
```diff
@@ -359,30 +359,80 @@
...
+            with the <a>service worker registration</a> associated with the <a>push subscription</a>
+            as <var>registration</var>, a <a>PushSubscription</a> instance representing the initial
+            <a>push subscription</a> as <var>oldSubscription</var> and a <a>PushSubscription</a>
+            instance representing the new <a>push subscription</a> as <var>newSubscription</var>.
+          </p>
+          <p>
+            <a>User agents</a> MAY continue to accept messages for the old <a>push subscription</a>
+            for a brief amount of time, but MUST stop doing so once the first message for the
+            refreshed <a>push subscription</a> has been received, as this indicates that the
+            <a>application server</a> has propagated the subscription change.
```
martinthomson: That creates an interesting externality: if the application is required to update two data sources with updated subscription details, it has to ensure that one source isn't used before the other is emplaced.  I think that's a reasonable global trade-off though.

I would restructure this to more directly link the MAY to the reason:

`To allow for time to propagate changes to <a>application servers</a>, a <a>user agent</a> MAY continue to accept messages for an old <a>push subscription</a> for a brief time after a refresh.  Once messages have been received for a refreshed <a>subscription</a>, any old <a>subscription</a> MUST be <a>deactivated</a>.`

index.html
```diff
@@ -423,8 +473,13 @@
         acquiring permission or determining the permission status.
       </p>
       <p>
-        When a permission is revoked, all <a>push subscriptions</a> created with that permission
-        MUST be <a>deactivated</a>.
+        When a permission is revoked, the <a>user agent</a> MAY
+        <a>fire the pushsubscriptionchange event</a> for subscriptions created with that permission,
+        with the <a>servce worker registration</a> associated with the <a>push subscription</a> as
```
martinthomson: typo: servce

index.html
```diff
@@ -423,8 +473,13 @@
...
+        <var>registration</var>, a <a>PushSubscription</a> instance representing the
+        <a>push subscription</a> as <var>oldSubscription</var>, and <code>null</code> as
+        <var>newSubscription</var>. The <a>user agent</a> MUST, in parallel, <a>deactivate</a>
```
martinthomson: the user agent must deactivate the affected subscriptions in parallel

index.html
```diff
@@ -1147,59 +1205,81 @@
...
+          <var>newSubscription</var>.
           </li>
-          <li>Invoke the <a>Handle Functional Event</a> algorithm with a <a>service worker
-          registration</a> of <var>registration</var> and <var>callbackSteps</var> set to the
-          following steps:
+          <li>Set the <code>oldSubscription</code> attribute of <var>event</var> to
+          <var>oldSubscription</var>.
+          </li>
+          <li>Invoke the <a>Handle Functional Event</a> algorithm with <var>event</var>,
+          <var>registration</var> and <var>callbackSteps</var> set to the following steps:
```
martinthomson: The list construction implies that all three inputs are set to the following steps.  I don't know how to improve that with some more restructuring though.

index.html
```diff
@@ -1147,59 +1205,81 @@
...
+          </p>
+          <p>
+            The <dfn>oldSubscription</dfn> attribute contains the details of the
+            <a>push subscription</a> that SHOULD NOT be used anymore. The value will be
+            <code>null</code> when the <a>user agent</a> is not able to provide the full set of
+            details, for example because of partial database corruption.
+          </p>
+          <p class="issue">
+            Should we include an attribute that indicates whether the <code>oldSubscription</code>
+            has been decisively invalidated? How would developers use this information?
```
martinthomson: I think that the answer here is no.  The intent of having the old subscription respond to push messages is to avoid races.  We should not encourage use of subscriptions once they are considered old.

index.html
```diff
@@ -359,30 +359,80 @@
...
+            <a>User agents</a> MAY continue to accept messages for the old <a>push subscription</a>
+            for a brief amount of time, but MUST stop doing so once the first message for the
+            refreshed <a>push subscription</a> has been received, as this indicates that the
+            <a>application server</a> has propagated the subscription change.
+          </p>
+
+          <p class="issue">
+            The steps for creating a subscription should be separated from the
+            <a lt="PushManager.subscribe">subscribe</a> method into an algorithm of its own so that
+            it can be referenced here. Similarly, the <a>PushSubscriptionOptions</a> provided to the
```
martinthomson: I would raise the issue on github.

martinthomson (approved):
This is definitely an improvement.

@kitcambridge, does this make sense to you?

index.html
ghost: Should we require the user agent to retry if the refresh is unsuccessful? (Or fire `pushsubscriptionchange` with `newSubscription` set to null, as a last resort?)

index.html
```diff
@@ -423,8 +473,13 @@
         acquiring permission or determining the permission status.
       </p>
       <p>
-        When a permission is revoked, all <a>push subscriptions</a> created with that permission
-        MUST be <a>deactivated</a>.
+        When a permission is revoked, the <a>user agent</a> MAY
```
ghost, line 477: What about if permission is granted again, after being revoked? @beverloo, I think you mentioned Chrome won't fire a `pushsubscriptionchange` event in that case, right?

index.html
ghost: I agree. Allowing the old subscription to briefly receive messages is helpful, but I don't think we want folks depending on it.

ghost:
Sorry for the delay. Modulo a couple of questions, this LGTM!

index.html
beverloo: Done

index.html
beverloo: I added the following paragraph:

> If the user agent is not able to refresh the push subscription, it should periodically retry the refresh. When the push subscription can no longer be used, for example because it has expired, the user agent must fire the pushsubscriptionchange event with the service worker registration associated with the push subscription as registration, a PushSubscription instance representing the deactivating push subscription as oldSubscription and null as the newSubscription.

I like Kit's suggestion of setting `newSubscription` to `null` since it gives the developer _something_.

index.html
beverloo: Thanks! Adopted.

index.html
beverloo: https://github.com/w3c/push-api/issues/235. Removed here.

index.html
beverloo: I actually think we found a case where we need to do this too. Let me file a more elaborate issue on it in a bit.

index.html
beverloo: Fixed.

index.html

index.html
beverloo: Slightly rephrased.

index.html
beverloo: Agreed. Thanks - removed.

beverloo (commented):
Thank you for the feedback. Let's add this and build upon it.

PR #237, opened 2017-02-17
Restructure the `push` event

beverloo:
This PR separates out the definition of the `push` event from the PushEvent interface and the action to take when receiving a message.

There's two changes that I'll point out specifically, the rest is moving things around.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1080,129 +1080,155 @@ navigator.serviceWorker.register('serviceworker.js').then(
       </section>
       <section>
         <h2>
-          <dfn>PushEvent</dfn> interface
+          The <dfn>push</dfn> Event
         </h2>
         <p>
-          The <a>PushEvent</a> interface represents a received <a>push message</a>.
+          The <a>push</a> event indicates that a <a>push message</a> has been received for a
+          <a>push subscription</a>.
         </p>
-        <pre class="idl">
-          typedef (BufferSource or USVString) PushMessageDataInit;
-
-          dictionary PushEventInit : ExtendableEventInit {
-            PushMessageDataInit data;
-          };
-
-          [Constructor(DOMString type, optional PushEventInit eventInitDict), Exposed=ServiceWorker]
-          interface PushEvent : ExtendableEvent {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1080,129 +1080,155 @@
...
+            byte sequence</a> from the "<code>data</code>" member of <var>eventInitDict</var>.
+            </li>
+            <li>Set the <code>data</code> attribute of the event to a new <a>PushMessageData</a>
+            instance with <code>bytes</code> set to <var>b</var>.
+            </li>
+          </ol>
+          <p>
+            The <dfn>data</dfn> attribute contains the data included in the <a>push message</a> when
+            included and the <a>user agent</a> verified its authenticity. The value will be set to
+            <code>null</code> in all other cases.
```
beverloo, line 1149: ^^^ This still is somewhat redundant (since `Receiving a Push Message` does the validation) and doesn't capture the case where a `PushEvent` is created from a `PushEventInit` structure. Would the following work for you?

> The \<dfn>data\</dfn> attribute contains the data associated with the \<a>push message\</a>. The value will be set to \<code>null\</code> when no data has been given.

index.html
```diff
@@ -1080,129 +1080,155 @@
...
+                  not successfully decrypted using the key pair associated with the <a>push
+                  subscription</a>.
+                </li>
+                <li>Let <var>data</var> be a new <a>PushMessageData</a> instance with the decrypted
+                plain text of the <a>push message</a>.
+                </li>
+              </ol>
+            </li>
+            <li><a>Fire the push event</a> with <var>registration</var> and <var>data</var>, and
+            set <var>event</var> to the returned <a>trusted event</a>.
```
beverloo, line 1199: ^^^ I removed the `If the push message contains valid data...` phrasing here since `data` would be `null` if it doesn't.

index.html
martinthomson: Yes, that's fine.

index.html
martinthomson: OK.

martinthomson (approved):

PR #238, opened 2017-02-17
Connectivity is not required to unsubscribe a subscription

### LINKED ISSUE #122: Allow UAs to unsubscribe offline "successfully"
The [unsubscribe algorithm](https://w3c.github.io/push-api/#widl-PushSubscription-unsubscribe-Promise-boolean) currently says:

> If it was not possible to access the push service, reject promise with a "NetworkError" exception and terminate these steps.

But it would be a nicer developer experience if unsubscribe were to immediately clear local state\* and resolve the promise successfully, then retry automatically to tell the push server about the unsubscription with exponential backoff (which may eventually give up).

This is what UAs must implement anyway if the website unregisters their Service Worker, since that will succeed irrespective of whether the device is offline.

Let's either require this in the spec, or change the language to allow a UA to do this, whilst permitting other UAs to still return NetworkErrors if they prefer that.

*: by clear local state, I mean make sure that `getSubscription()` resolves as `null`, and push messages are no longer delivered to the Service Worker. Until the unregistration is successfully transmitted to the push server, the push server will still consider it valid (which the developer may be able to observe).

beverloo:
Fixes #122

### DIFF
--- a/index.html
+++ b/index.html
@@ -251,7 +251,6 @@
         "http://www.w3.org/TR/dom/#invalidcharactererror"><dfn>InvalidCharacterError</dfn></a></code>,
         <code><a href=
         "http://www.w3.org/TR/dom/#securityerror"><dfn>SecurityError</dfn></a></code>,
-        <code><a href="http://www.w3.org/TR/dom/#networkerror"><dfn>NetworkError</dfn></a></code>,
         <a href="http://www.w3.org/TR/dom/#concept-event-listener"><dfn>event listener</dfn></a>,
         and <a href="http://www.w3.org/TR/dom/#concept-event-constructor"><dfn>steps for
         constructing events</dfn></a> are defined in [[!DOM]].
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -907,13 +906,17 @@
...
-        </li>
-        <li>If it was not possible to access the <a>push service</a>, reject <var>promise</var>
-        with a "<a>NetworkError</a>" exception and terminate these steps.
+        <li>Run the following step in parallel:
+          <ol>
+            <li><a>Deactivate</a> the <a>push subscription</a>. The <a>user agent</a> MUST NOT
+            deliver any further <a>push messages</a> for the <a>push subscription</a>.
+            <p>If the <a>user agent</a> failed to request the <a>push service</a> to <a>deactivate</a>
+            the <a>push subscription</a>, for example because of network failures, it SHOULD retry
+            the request to the <a>push service</a> for a reasonable amount of time.
```
martinthomson, line 915: Do we want to hold the resolution of the promise on this concluding, or are we really just doing this as a background task.  If this is a background task, then the browser could resolve the promise after whatever local steps it might need to take have been taken.  For instance, Firefox would remove entries from indexedDB, but then the network actions could happen in their own time.

martinthomson (approved):

index.html
beverloo: Yes. My intention (with the "run the following step in parallel" sub-step) is that we hold the promise from resolving until the necessary work to stop messages from being delivered to the Service Worker has been completed, but not any longer than that (i.e. network activity to the push service).

Issue #243, opened 2017-03-15
Non-Browser Vendor Supplied Push Server Support


msporny:
Hi all,

We (Digital Bazaar, W3C Member company) will soon be using the Web Push API in production, so first off, thanks for this feature/functionality. It's great and enabled us to deploy our digital wallet products to mobile using the Web Platform instead of native apps.

I've read through every issue that I could find relating to push service registration and note that the push service that one uses effectively boils down to either Google's servers or Mozilla's servers. While I understand the desire to get this deployed in a way that's easy for Web Developers, that the Push Notifications go through servers that we don't control is a problem for some of our more security conscious customers. We do realize that payloads can be encrypted, but we'd rather just run a Web Push service ourselves (again, it ticks a box for our customers).

My understanding is that there is no intention to support a more decentralized model for the Web Push servers and one MUST depend on the centralized Google or Mozilla servers. I haven't been involved in Web Standards for long, just over a decade, but I can't think of any other Web Platform spec that requires proprietary services (other than EME) to operate.

Has the thinking of this changed lately? If so, is there a timeline for supporting non-browser vendor supplied Push server services?

beverloo:
Hi Manu, exciting to hear that you'll soon be rolling out support for Web Push!

Mobile devices were the primary motivation for the Push API. They come with a series of unique constraints, notably that they're battery powered. They also usually come with built-in push services that are available to every application installed on the device. By enabling multiple applications to share a connection instead of each establishing their own, a very substantial amount of power can be saved. (Networking, specifically the radio on a mobile device, tends to be the largest contributor to battery usage after the screen.)

The Push API exists to enable web applications to benefit from this connection.

We've decided against supporting a bring-your-own-push-service model because establishing multiple connections goes against the goals of the Push API. It has a significant negative impact on the user (by draining their battery faster), and is, in fact, very hard to impossible to support on various mobile platforms.

As you note, this comes with important privacy and security considerations. For that reason the Push API goes hand-in-hand with the Web Push Protocol (RFC8030) standardized at the IETF. It specifically designates the intermediaries —the push services— as untrusted. In effect, we require end-to-end encryption for messages, and give you some other tools such as [message padding](https://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-08#section-2) to even hide things like the length of your payload.

martinthomson:
This is a very common question.  Peter captured this well.  RFC 8030 is careful to describe the same constraints.

msporny:
[...]

PR #244, opened 2017-03-26
Add note about single push service per user agent.

msporny:
This is a follow up to https://github.com/w3c/push-api/issues/243. The PR adds a note explaining that there is only one push service per user agent and provides some reasoning for the design decision. Our engineers spent a fair bit of time trying to understand if there was a way to change the push service and this text would have saved them some time.

### DIFF
--- a/index.html
+++ b/index.html
@@ -444,6 +444,13 @@
           push service serves the <a>push endpoint</a> or <a data-lt="push endpoints">endpoints</a>
           for the <a>push subscriptions</a> it serves.
         </p>
+        <p>
+          There is only one <a>push service</a> per <a>user agent</a> and it cannot be changed
+          from the default value. This limitation is due to a variety of performance-related
+          concerns, including the complexity of running reliable <a>push services</a> and the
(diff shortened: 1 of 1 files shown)


msporny:
Marked as non-substantive for IPR from ash-nazg.

Err... should I have been allowed to do that? I think it's editorial... and our company joined the group... so we're good either way.

buhrmi:
Hello,

I stumbled over this too because I was under the impression that it would be easy to change the push server that is being used. A hint like that would have saved me time.

Also, it should be made possible for application developers to tell the user agent which push server to use. How would I go about requesting this?

Stefan

beverloo:
@msporny - missed this, thanks! Let's merge it.

@buhrmi - you could file an issue, but we've repeatedly explained why we're not planning to support this. See e.g. https://github.com/w3c/push-api/issues/243#issuecomment-286782918 which I linked you to earlier today.

buhrmi:
Indeed. And while I whole-heartedly understand the explanations, I'm not really satisfied by them...

PR #245, opened 2017-03-28
Make use of quotes consistent in spec.

msporny:
This just fixes a few minor quote consistency issues in the spec.


msporny:
Marked as non-substantive for IPR from ash-nazg.

PR #248, opened 2017-04-11
Add PushSubscription.expirationTime

### LINKED ISSUE #86: PushSubscription should have an attribute for ExpirationTime
As discussed on the mailing list [1], it'd be very helpful if PushRegistration can have a read-only attribute for ExpirationTime.  Webapps can be more proactive if the ExpirationTime is set.

partial interface PushRegistration {
    readonly  attribute  Date?  expirationTime;
};

[1] http://lists.w3.org/Archives/Public/public-webapps/2014OctDec/0227.html

beverloo:
The expiration time of a push subscription indicates the time, in
milliseconds since the UNIX epoch (UTC), at which the subscription will
be deactivated.

This is an important addition for user agents that refresh their
subscriptions, as it gives the developer knowledge of a final time after
which they can garbage collect the stored subscriptions if the user has
not communicated the details of the refreshed subscription.

Fixes #86

### DIFF
--- a/index.html
+++ b/index.html
@@ -293,17 +293,19 @@
         [[!FILEAPI]].
       </p>
       <p>
-        <code><a href="http://www.w3.org/TR/WebIDL/#idl-any"><dfn>Any</dfn></a></code>,
+        <code><a href="https://www.w3.org/TR/WebIDL/#idl-any"><dfn>Any</dfn></a></code>,
         <code><a href=
-        "http://heycam.github.io/webidl/#idl-ArrayBuffer"><dfn>ArrayBuffer</dfn></a></code>,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -365,6 +367,12 @@
           subscription</a>.
         </p>
         <p>
+          A <a>push subscription</a> MAY have an associated <dfn>subscription expiration time</dfn>.
+          When set, it MUST be the time, in milliseconds since 00:00:00 UTC on 1 January 1970,
+          at which the subscription will be <a>deactivated</a>. The <a>user agent</a> SHOULD attempt
+          to <a>refresh</a> to push subscription before the subscription expires.
```
martinthomson: s/to/the

martinthomson (approved):

index.html
beverloo: Done - thanks!

PR #252, opened 2017-04-21
Add PushManager.supportedContentEncodings

### LINKED ISSUE #251: User Agents should report accepted content-encoding types.
Recently, a bug filed against a webpush subscription library highlighted a shortcoming.

https://github.com/web-push-libs/web-push-php/issues/48#issuecomment-295416292

Currently, there are two in production encryption encoding types, "aesgcm" and "aes128gcm". The "voice of authority" about what types of accepted content types is the UA. The sorts of allowed encryption is not communicated to the subscription update provider.

I would like to propose that the returned [PublishSubscription](https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription) object  `options` object be modified to include a "encodingtypes" list of allowed ECE encoding types. (e.g. `["aesgcm", "aes128gcm"]`) This method would also allow future content types to be relayed. If no "encodingtypes" field is present, then the provider must assume "aesgcm" encoding, to allow for older UAs.

This field would also help indicate "updated" UAs which can take advantage of the newer draft specifications.

(Some additional, albeit incorrectly located, [discussion is here](https://mailarchive.ietf.org/arch/msg/webpush/uDR2PzpChfn24TwNWZBfR6fuKp0).)

beverloo:
Fixes #251

### DIFF
--- a/index.html
+++ b/index.html
@@ -314,6 +314,10 @@
         specification assumes the use of this protocol; alternative protocols are expected to
         provide compatible semantics.
       </p>
+      <p>
+        The <dfn>Content-Encoding</dfn> HTTP header, described in Section 3.1.2.2 of [[!RFC7231]],
+        indicates the content coding applied to the payload of a <a>push message</a>.
+      </p>
     </section>
     <section>
       <h2>
@@ -642,11 +646,24 @@ navigator.serviceWorker.register('serviceworker.js').then(
       </p>
       <pre class="idl">
         interface PushManager {
+          [SameObject] static readonly attribute FrozenArray&lt;DOMString&gt; supportedContentEncodings;
+
           Promise&lt;PushSubscription&gt; subscribe(optional PushSubscriptionOptionsInit options);
           Promise&lt;PushSubscription?&gt; getSubscription();
           Promise&lt;PushPermissionState&gt; permissionState(optional PushSubscriptionOptionsInit options);
         };
       </pre>
+      <p>
+        The <dfn>supportedContentEncodings</dfn> attribute exposes the sequence of supported content
+        codings that can be used to encrypt the payload of a <a>push message</a>. A content coding
+        is indicated using the <a>Content-Encoding</a> header field when requesting the sending of
+        a <a>push message</a> from the <a>push service</a>.
+      </p>
+      <p>
+        <a>User agents</a> MUST support the <code>aes128gcm</code> content coding defined in
+        [[!WEBPUSH-ENCRYPTION]], and MAY support content codings defined in previous versions of
+        the draft for compatibility reasons.
+      </p>
       <p>
         The <dfn>subscribe</dfn> method when invoked MUST run the following steps:
       </p>


beverloo:
I'm sure the phrasing can be improved, @martinthomson. Most open to suggestions!

index.html
```diff
@@ -314,6 +314,10 @@
         specification assumes the use of this protocol; alternative protocols are expected to
         provide compatible semantics.
       </p>
+      <p>
+        The <dfn>Content-Encoding</dfn> HTTP header, described in section 3.1.2.2. of [[!RFC7231]],
```
martinthomson: The RFC style guide references sections using "Section 3.1.2.2", capital on Section and no trailing period.

index.html
```diff
@@ -642,12 +646,25 @@
       </p>
       <pre class="idl">
         interface PushManager {
+          [SameObject] static readonly attribute FrozenArray&lt;DOMString&gt; supportedContentEncodings;
```
martinthomson, line 649: Hmm, it's called a content coding, but the header is content-encoding.  I think that `supportedContentEncodings` is probably the best choice, because other than us nerds who work on this stuff, the difference is meaningless.

index.html
```diff
@@ -642,12 +646,25 @@
...
+
           Promise&lt;PushSubscription&gt; subscribe(optional PushSubscriptionOptionsInit options);
           Promise&lt;PushSubscription?&gt; getSubscription();
           Promise&lt;PushPermissionState&gt; permissionState(optional PushSubscriptionOptionsInit options);
         };
       </pre>
       <p>
+        The <dfn>supportedContentEncodings</dfn> attribute exposes the sequence of supported content
+        codings using which the payload of a <a>push message</a> can be encrypted. When a content
```
martinthomson: "... codings that can be used to encrypt the payload of a push message."

index.html
```diff
@@ -642,12 +646,25 @@
...
+        coding is used, it MUST be set in the <a>Content-Encoding</a> HTTP header when sending a
+        <a>push message</a> to the <a>push service</a>.
```
martinthomson: I don't think that you want the MUST requirement here.

"A content coding is indicated using the <a>Content-Encoding</a> header field when requesting the sending of a <a>push message</a> from the <a>push service</a>."   ...Maybe.

martinthomson (changes requested):

martinthomson (approved):

index.html

index.html
beverloo: Agreed - the developer is exposed to the Content-Encoding header more than the fact it accepts a (list of) content coding.

index.html

index.html
beverloo: Done. I considered changing it around (sending from the AS) but there's other cases too, so I adopted your suggestion.

beverloo (commented):
Thanks!

Issue #232, opened 2016-12-14
Final IDL cleanup


martinthomson:
See https://github.com/w3c/push-api/pull/231#issuecomment-265552037

martinthomson:
I think that we're in reasonable shape.  Let's continue to play whack-a-mole as we notice errors.

PR #253, opened 2017-04-21
Fix a series of warnings in the specification

beverloo:
What remains are nine warnings about missing definitions in the *Init dictionaries, which would be either duplicate or refer to the definitions of the identically named properties in the interfaces.

I don't think it's a fantastic idea to do either. @marcoscaceres would you know what the idiomatic way of dealing with those warnings is? A few examples:

  * ``No <dfn> for userVisibleOnly in PushSubscriptionOptionsInit. Please define it and link to spec that declares it. See https://github.com/w3c/respec/wiki/data--cite``
  * ``No <dfn> for data in PushEventInit. Please define it and link to spec that declares it. See https://github.com/w3c/respec/wiki/data--cite``

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,8 +5,7 @@
       Push API
     </title>
     <meta charset="utf-8">
-    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove">
-    </script>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
     <script class="remove">
     // (this is to make tidy happy)
       // See http://www.w3.org/respec/ for ReSpec documentation.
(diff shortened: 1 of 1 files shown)


beverloo:
That question basically is #232 by the way, given your comment on https://github.com/w3c/push-api/pull/231#issuecomment-265552037. Happy to make the change of course.

marcoscaceres:
I'm on vacation till next week. If it's ok, will take a look then. However, you can see "best practice"(tm) in the Payment Request spec.

beverloo:
Cool, thanks! I'll take a look at the Payment Request specification to see how they solve this. The rest of the changes in this PR are editorial, so I'll go forward and submit this.

PR #254, opened 2017-04-21
Separate out the steps for creating a subscription

### LINKED ISSUE #235: Separate out an algorithm for creating a push subscription
This is currently only done when the developer calls `subscribe()`, but now that user agents can refresh subscriptions we'll have to reuse it.

beverloo:
Fixes #235. This builds on #253, the actual changes are included in the following commit:

https://github.com/w3c/push-api/pull/254/commits/56fa7fccf3059e60fa5b21fac4260bbb6e293139

### DIFF
--- a/index.html
+++ b/index.html
@@ -386,6 +386,35 @@
           <var>oldSubscription</var> and a <a>PushSubscription</a> instance representing the
           <a>push subscription</a> having the new keys as <var>newSubscription</var>.
         </p>
+        <p>
+          To <dfn>create a push subscription</dfn>, given an <a>PushSubscriptionOptions</a> object
+          of <var>options</var>, the <a>user agent</a> must run the following steps:
+        </p>
+        <ol>
+          <li>Let <var>subscription</var> be a new <a>push subscription</a>.
+          </li>
+          <li>Set the <code>options</code> attribute of <var>subscription</var> to be a copy of
+          <var>options</var>.
+          </li>
+          <li>Generate a new P-256 <a>ECDH</a> key pair [[!X9.62]]. Store the private key in an
+          internal slot on <var>subscription</var>; this value MUST NOT be made available to
+          applications. The public key is also stored in an internal slot and can be retrieved by
+          calling the <code>getKey</code> method of the <a>PushSubscription</a> with an argument of
+          <a data-link-for="PushEncryptionKeyName">p256dh</a>.
+          </li>
+          <li>Generate a new authentication secret, which is a sequence of octets as defined in
+          [[!WEBPUSH-ENCRYPTION]]. Store the authentication secret in an internal slot on
+          <var>subscription</var>. This key can be retrieved by calling the <code>getKey</code>
+          method of the <a>PushSubscription</a> with an argument of <a data-link-for=
+          "PushEncryptionKeyName">auth</a>.
+          </li>
+          <li>Make a request to the <a>push service</a> to create a new <a>push subscription</a>.
+          Include the <a data-link-for="PushSubscriptionOptions">applicationServerKey</a> attribute
+          of <var>options</var> when it has been set.
+          </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -383,6 +382,35 @@
...
+          To <dfn>create a push subscription</dfn>, given an <a>PushSubscriptionOptions</a> object
+          of <var>options</var>, the <a>user agent</a> must run the following steps:
+        </p>
+        <ol>
+          <li>Let <var>subscription</var> be a new <a>push subscription</a>.
+          </li>
+          <li>Set the <code>options</code> attribute of <var>subscription</var> to be a copy of
+          <var>options</var>.
+          </li>
+          <li>Generate a new P-256 <a>ECDH</a> key pair. Store the private key in an internal slot
```
martinthomson: Cite X9.62 on the first sentence.

index.html
```diff
@@ -383,6 +382,35 @@
...
+          on <var>subscription</var>; this value MUST NOT be made available to applications. The
+          public key is also stored in an internal slot and can be retrieved by calling the <code>
+            getKey</code> method of the <a>PushSubscription</a> with an argument of
```
martinthomson: Is this indent what tidy produces?  Strange.

index.html
```diff
@@ -383,6 +382,35 @@
...
+          </li>
+          <li>Generate a new authentication secret, which is a sequence of octets as defined in
+          [[!WEBPUSH-ENCRYPTION]]. Store the authentication secret in an internal slot on
+          <var>subscription</var>. This key can be retrieved by calling the <code>getKey</code>
+          method of the <a>PushSubscription</a> with an argument of <a data-link-for=
+          "PushEncryptionKeyName">auth</a>.
+          </li>
+          <li>Make a request to the <a>push service</a> to create a new <a>push subscription</a>.
+          Include the <a data-link-for="PushSubscriptionOptions">applicationServerKey</a> attribute
+          of <var>options</var> when it contains a valid point on the P-256 curve.
```
martinthomson: We reject the call if `applicationServerKey` is invalid, so you can remove the "when it contains a valid point" bit.

index.html
```diff
@@ -868,9 +885,14 @@
         is one, or <code>null</code> otherwise.
       </p>
       <p>
+        When getting the <dfn>options</dfn> attribute, the <a>user agent</a> MUST return a
+        <code><a>PushSubscriptionOptions</a></code> object representing the options associated with
```
martinthomson: s/object/dictionary ?

index.html
```diff
@@ -930,7 +953,7 @@
         </li>
       </ol>
       <p>
-        The serializer for a <a>PushSubscription</a> invokes the following steps:
+        The <dfn>serializer</dfn> for a <a>PushSubscription</a> invokes the following steps:
```
martinthomson: no data-dfn-for here?

martinthomson (approved):

index.html
beverloo: Yes, presumably because the \<code> block is open?

index.html
beverloo: "when it contains a valid point on the P-256 curve" -> "when it has been set"

index.html
beverloo: No, because PSO is the object, PSOInit is the dictionary. We return PSO.

index.html
beverloo: The owning \<section> has been annotated with the `data-dfn-for` attribute, making that redundant. That doesn't work when a section contains multiple identically named attributes, as is the case with the Init dictionaries.

index.html
beverloo: I'm not sure how to phrase this. "as described in [[!X9.62]]"?

The actual curve parameters are defined in Section 10.2, should we refer that too?

beverloo: Went with just [[!X9.62]] since "as described in" doesn't add much either.

PR #255, opened 2017-05-03
Define exposure of the Push API interfaces

beverloo:
https://w3c.github.io/ServiceWorker/v1/#serviceworkerregistration-interface defines this for `ServiceWorkerRegistration`, this is what UAs implement, so we should make sure our interfaces match.

### DIFF
--- a/index.html
+++ b/index.html
@@ -672,6 +672,7 @@ navigator.serviceWorker.register('serviceworker.js').then(
         The <a>PushManager</a> interface defines the operations to access <a>push services</a>.
       </p>
       <pre class="idl">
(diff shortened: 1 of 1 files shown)


beverloo:
For the record, `[SecureContext]` is also used for `ServiceWorkerRegistration`, but we'll have to assess compatibility because today we expose our interfaces in non-secure documents as well.

martinthomson:
Let's track `[SecureContext]` separately.  We probably want to get there, since push doesn't really work without service workers, but if you need to poke at a few things, then you should do that.

PR #262, opened 2017-06-16
Add a list of authentication schemes

martinthomson:
We added the ability to discover the content encodings, but more or less forgot to add the supported authentication schemes.  This corrects that oversight.

### DIFF
--- a/index.html
+++ b/index.html
@@ -675,6 +675,7 @@ navigator.serviceWorker.register('serviceworker.js').then(
         [Exposed=(Window,Worker)]
         interface PushManager {
           [SameObject] static readonly attribute FrozenArray&lt;DOMString&gt; supportedContentEncodings;
+          [SameObject] static readonly attribute FrozenArray&lt;DOMString&gt; supportedAuthenticationSchemes;
 
           Promise&lt;PushSubscription&gt; subscribe(optional PushSubscriptionOptionsInit options);
           Promise&lt;PushSubscription?&gt; getSubscription();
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -688,6 +689,11 @@
         when requesting the sending of a <a>push message</a> from the <a>push service</a>.
       </p>
       <p>
+        The <dfn data-dfn-for="PushManager">supportedAuthenticationSchemes</dfn> attribute exposes
+        the sequence of supported authentication schemes that can be used by <a>applications
```
beverloo, line 693: nit: s/applications/application/

index.html
```diff
@@ -688,6 +689,11 @@
...
+        servers</a> to authenticate with the <a>push service</a>.
+      </p>
```
beverloo, line 695: Can we list at least "vapid" (and ideally require it -- is that feasible?), but ideally also explain what sort of other values one might expect in this array? Would Chrome include a `x-google-gcm` or similar if a proprietary mechanism were available?

index.html
```diff
@@ -868,7 +874,11 @@
           key is used to generate an authentication token.
         </p>
         <p>
-          If present, the value of <a data-link-for=
+          If present, the value of the <code><a data-link-for=
+          "PushSubscriptionOptions">applicationServerKey</a></code> needs to be valid according to
+          one of the supported authentication schemes (see <code><a data-link-for=
+          "PushManager">supportedAuthenticationSchemes</a></code>). For the "vapid" scheme defined
+          in [[!WEBPUSH-VAPID]], the value of <a data-link-for=
```
beverloo, line 881: This now implies that `applicationServerKey` can contain non-P-256 values too. That creates a conflict in the algorithm at 8.5.2, and makes the paragraph just before this one ambiguous: it mentions _an elliptic curve_, but what about other sort of values?

index.html
beverloo: Also, does `vapid` refer to the latest version of the draft, or the previous one (still using `Crypto-Key`) that still has more implementations?

beverloo (commented):
Thanks! I'm a bit in two minds about this. On one hand I can see the value, the implementation cost _most likely_ is low and modulo a few more details it's simple enough.

On the other hand, the property describes capabilities of the push service, not the user agent. Since they can update independently, a UA is now expected to synchronise the available authentication schemes with the push service it has chosen. (Either though an occasional handshake or by aligning release schedules.) That's OK for us, but may not be equally feasible for everyone.

/cc @ShijunS FYI

martinthomson:
FWIW, I share your reservations here.

martinthomson:
I think that we can abandon this.  VAPID is looking to gain support for challenges, which in combination with the change to reject subscriptions if VAPID is required, mostly cover this off.

PR #264, opened 2017-06-20
Add answers to the Security and Privacy self-review

beverloo:
See https://www.w3.org/TR/security-privacy-questionnaire/. It seems easy enough to just check this in.

/cc @LJWatson

### DIFF
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -1,0 +1,99 @@
+# Security and Privacy Questionnaire for the Push API
+
+This document answers the questions raised in the
+[Security and Privacy Questionnaire](https://www.w3.org/TR/security-privacy-questionnaire/)
+in context of the [Push API](https://w3c.github.io/push-api/).
+
+## 3.1 Does this specification deal with personally-identifiable information?
+
+The Push API issues a subscription that has an endpoint unique to the user. It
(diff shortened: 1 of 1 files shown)


LJWatson:
Thanks @Beverloo. LGTM.

When this is merged, I'll ping the Security and Privacy IGs for a review (tracked via #265).

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,99 @@
...
+
+## 3.8 Does this specification allow an origin access to sensors on a user’s device?
+
+No.
+
+## 3.9 Does this specification allow an origin access to aspects of a user’s local computing environment?
+
+Yes, the push service selected by the device. In some cases, most commonly on
+desktop platforms, a user agent includes a push service client separate from the
+push service made available by the operating system.
```
martinthomson: I'm not sure what "access" means here.  The specification allows an origin to identify the push service.  I would say that the push service is selected by the user agent and that that might use the same push service as other applications.

martinthomson (approved):
Not sure about the interpretation of the one question, but I'm sure that this will do.

security-privacy-questionnaire.md
beverloo: Updated, thanks!

PR #267, opened 2017-06-22
Migrate the serializer to a toJSON method.

### LINKED ISSUE #266: WebIDL serializer has been deprecated in favor of toJSON operation
Hi!

We recently [deprecated WebIDL serializers][0]. You can now directly specify [toJSON operations][1] instead, which you previously weren't allowed to do.

To deal with common cases, we added a new [\[Default\] extended attribute][2] which triggers the [default toJSON operation][3] that behaves similarly to how `serializers={attributes}` or `serializers={attributes, inherit}` used to. That is, it serializes all attributes that are of a [JSON type][4] into a vanilla JSON object.

It seems at least the following interface in this spec is impacted by this change:

- [ ] [`PushSubscription`](https://w3c.github.io/push-api/#pushsubscription-interface)

There's [an example][5] of a custom `toJSON` operation in the WebIDL spec to get you going. Overall, it doesn't seem like you'll need to change a lot of things beyond specifying an output format more explicitly.

I'm sorry for the inconvenience this causes, but our hope is that this ultimately makes things a lot simpler and clearer for everybody.

Please feel free to reach out if you have any questions.

Thanks!

[0]: https://github.com/heycam/webidl/commit/a505f33
[1]: https://heycam.github.io/webidl/#idl-tojson-operation
[2]: https://heycam.github.io/webidl/#Default
[3]: https://heycam.github.io/webidl/#es-default-tojson
[4]: https://heycam.github.io/webidl/#dfn-json-types
[5]: https://heycam.github.io/webidl/#tojson-example

beverloo:
Fixes #266. How's this, @tobie?

***
[Preview](https://s3.amazonaws.com/pr-preview/beverloo/push-api/to-json.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/push-api/2b1d75a...beverloo:323d302.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -232,10 +232,12 @@
       </p>
       <p>
         <code><a href=
-        "https://tc39.github.io/ecma262/#sec-promise-objects"><dfn>Promise</dfn></a></code>, and
+        "https://tc39.github.io/ecma262/#sec-promise-objects"><dfn>Promise</dfn></a></code>,
         <code><a href=
-        "https://tc39.github.io/ecma262/#sec-json.parse"><dfn>JSON.parse</dfn></a></code> are
-        defined in [[!ECMASCRIPT]].
+        "https://tc39.github.io/ecma262/#sec-json.parse"><dfn>JSON.parse</dfn></a></code> and
+        <code><a href=
+        "https://tc39.github.io/ecma262/#sec-json.stringify"><dfn>JSON.stringify</dfn></a></code>
+        are defined in [[!ECMASCRIPT]].
       </p>
       <p>
         <code><a href="http://www.w3.org/TR/dom/#eventinit"><dfn>EventInit</dfn></a></code>,
@@ -301,10 +303,14 @@
         <code><a href=
         "https://heycam.github.io/webidl/#DOMTimeStamp"><dfn>DOMTimeStamp</dfn></a></code>,
         <code><a href=
+        "https://heycam.github.io/webidl/#dfn-json-types"><dfn>JSON type</dfn></a></code>,
+        <code><a href=
         "https://heycam.github.io/webidl/#notallowederror"><dfn>NotAllowedError</dfn></a></code>,
-        and <code><a href=
-        "https://heycam.github.io/webidl/#idl-USVString"><dfn>USVString</dfn></a></code> are
-        defined in [[!WEBIDL]].
+        <code><a href=
+        "https://heycam.github.io/webidl/#idl-USVString"><dfn>USVString</dfn></a></code> and
+        <code><a href=
+        "https://heycam.github.io/webidl/#get-the-underlying-value"><dfn>get the underlying value</dfn></a></code>
+        are defined in [[!WEBIDL]].
       </p>
       <p>
         The <dfn>web push protocol</dfn> [[!WEBPUSH-PROTOCOL]] describes a protocol that enables
@@ -896,12 +902,24 @@ navigator.serviceWorker.register('serviceworker.js').then(
           [SameObject] readonly attribute PushSubscriptionOptions options;
           ArrayBuffer? getKey(PushEncryptionKeyName name);
           Promise&lt;boolean&gt; unsubscribe();
-          serializer;
+
+          PushSubscriptionJSON toJSON();
+        };
+
+        dictionary PushSubscriptionJSON {
+          USVString endpoint;
+          DOMTimeStamp? expirationTime;
+
+          // TODO: This should be record&lt;DOMString, USVString&gt;.
+          // https://github.com/w3c/respec/issues/1297
+          record keys;
         };
       </pre>
       <p>
         When getting the <dfn>endpoint</dfn> attribute, the <a>user agent</a> MUST return the
-        <a>push endpoint</a> associated with the <a>push subscription</a>.
+        <a>push endpoint</a> associated with the <a>push subscription</a>. The <a>user agent</a>
+        MUST use a serialization method that does not contain input-dependent branchs (that is, one
+        that is constant time).
       </p>
       <p>
         When getting the <dfn>expirationTime</dfn> attribute, the <a>user agent</a> MUST return the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -977,22 +986,20 @@
...
         branchs (that is, one that is constant time). Note that a URL - as ASCII text - will not
         ordinarily require special treatment.
         </li>
-        <li>Add an entry to <var>map</var> whose key name is <code>expirationTime</code> and whose
-        value is the result of converting the <code>expirationTime</code> attribute to a serialized
-        value.
-        </li>
-        <li>Let <var>keys</var> be an empty map.
+        <li>Set the <code>expirationTime</code> property of <var>json</var> to the result of
+        converting the <a data-link-for="PushSubscription">expirationTime</a> attribute to a
```
tobie: I'd change that to something like the below:

```
Set the <code>expirationTime</code> property of <var>json</var>
to the result of <a data-link-for="get the underlying value">getting the underlying value</a>
of the <a data-link-for="PushSubscription">expirationTime</a> attribute,
given this PushSubscription object.
```

_(Note this references https://heycam.github.io/webidl/#get-the-underlying-value.)_

Additionally, you can use the <var>json</var>["expirationTime"] notation like so:

```
Set <var>json</var>["expirationTime"] to the result of […]
```

index.html
```diff
@@ -1009,20 +1016,36 @@
...
-        </li>
+        <li>Return <var>json</var>.
       </ol>
+
+      <p>
+        A <dfn>PushSubscriptionJSON</dfn> dictionary represents the <a>JSON type</a> of a
+        <a>PushSubscription</a> that can be converted into a JSON string.
+      </p>
+      <p>
+        The <dfn data-dfn-for="PushSubscriptionJSON">endpoint</dfn> contains the serialized value of
```
tobie: I'm uneasy about using the serialized value term, as it's not defined anywhere. I'd use the "underlying value" here and again reference https://heycam.github.io/webidl/#get-the-underlying-value.

index.html
```diff
@@ -896,7 +898,14 @@
...
           Promise&lt;boolean&gt; unsubscribe();
-          serializer;
+
+          PushSubscriptionJSON toJSON();
+        };
+
+        dictionary PushSubscriptionJSON {
+          USVString endpoint;
+          DOMTimeStamp? expirationTime;
+          record keys;
```
tobie: So this should probably be `record<USVString, USVString>` (or `DOMString`?—I must admit still being unsure when to use which.)

index.html
```diff
@@ -977,22 +986,20 @@
...
         </li>
-        <li>Add an entry to <var>map</var> whose key name is <code>expirationTime</code> and whose
-        value is the result of converting the <code>expirationTime</code> attribute to a serialized
-        value.
-        </li>
-        <li>Let <var>keys</var> be an empty map.
+        <li>Set the <code>expirationTime</code> property of <var>json</var> to the result of
+        converting the <a data-link-for="PushSubscription">expirationTime</a> attribute to a
+        serialized value.
         </li>
```
tobie: Here's I write something like this:

```
Let <var>keys</var> be a new empty instance of record<USVString, USVString>.
```

index.html
```diff
@@ -1009,20 +1016,36 @@
             <var>b</var> as a <a>USVString</a>. The <a>user agent</a> MUST use a serialization
             method that does not branch based on the value of <var>b</var>.
             </li>
-            <li>Add an entry to <var>keys</var> whose key name is the name of <var>i</var> and
-            whose value is <var>s</var>.
+            <li>Add an entry to the <a data-link-for="PushSubscriptionJSON">keys</a> property of
```
tobie: And here:

```
Set <var>keys</var>[<var>i</var>] to <var>s</var>.
```

index.html
```diff
@@ -1009,20 +1016,36 @@
...
+            <var>json</var> whose key is the name of <var>i</var> and whose value is <var>s</var>.
             </li>
           </ol>
         </li>
```
tobie: And finally:

```
Set <var>json</var>["keys"] to <var>keys</var>.
```

index.html
```diff
@@ -1009,20 +1016,36 @@
...
+      <p>
+        The <dfn data-dfn-for="PushSubscriptionJSON">endpoint</dfn> contains the serialized value of
+        the <a data-link-for="PushSubscription">endpoint</a> attribute.
+      </p>
+      <p>
+        The <dfn data-dfn-for="PushSubscriptionJSON">expirationTime</dfn> contains the serialized
+        value of the <a data-link-for="PushSubscription">expirationTime</a> attribute.
+      </p>
+      <p>
+        The <dfn data-dfn-for="PushSubscriptionJSON">keys</dfn> contains a mapping from each
```
tobie: I'd replace the above by:

```
The <dfn data-dfn-for="PushSubscriptionJSON">keys</dfn> record contains an entry for each
```
(see https://infra.spec.whatwg.org/#map-entry if you want to reference the term properly.)

index.html
```diff
@@ -1009,20 +1016,36 @@
...
-        <var>keys</var>.
-        </li>
-        <li>Return <var>map</var>.
-        </li>
+        <li>Return <var>json</var>.
       </ol>
+
+      <p>
+        A <dfn>PushSubscriptionJSON</dfn> dictionary represents the <a>JSON type</a> of a
+        <a>PushSubscription</a> that can be converted into a JSON string.
```
tobie: (and eventually append: "by JSON.stringify.")

tobie (approved):
So this is the first conversion from a custom serializer in the wild, so bear with me and please feel free to ask for clarification if any of my comments require it.

Overall this looks good.

Added a couple of nits and proposals to shorten the prose using [infra](https://infra.spec.whatwg.org/) language which WebIDL now normatively references.

tobie:
@domenic, do you want to give my comments a quick second look to see if we're in agreement, here?

index.html
beverloo: Ah yeah, I forgot to follow up on that - ReSpec seems to choke on `record<K, V>` syntax for some reason. The key should be a `DOMString` since it's one of the `PushEncryptionKeyName` values, and enum values are defined to be `DOMString`s.

index.html
beverloo: Done. Also for `endpoint`. I like that `get the underlying value` avoids duplication.

index.html
beverloo: Done, but with a `DOMString`.

index.html

index.html

index.html
```diff
@@ -977,22 +986,20 @@
...
       <ol>
-        <li>Let <var>map</var> be an empty map.
+        <li>Let <var>json</var> be a new <code>PushSubscriptionJSON</code> dictionary.
         </li>
-        <li>Add an entry to <var>map</var> whose key name is <code>endpoint</code> and whose value
-        is the result of converting the <code>endpoint</code> attribute to a serialized value. The
+        <li>Set the <code>endpoint</code> property of <var>json</var> to the result of converting
+        the <a data-link-for="PushSubscription">endpoint</a> attribute to a serialized value. The
         <a>user agent</a> MUST use a serialization method that does not contain input-dependent
         branchs (that is, one that is constant time). Note that a URL - as ASCII text - will not
```
domenic: Preexisting text comment: URLs are not in the general case ASCII...

index.html

index.html
tobie: Heh! :)

index.html
beverloo: Agreed, done.

index.html
```diff
@@ -1009,20 +1016,36 @@
...
-        <li>Add an entry to <var>map</var> whose key name is <code>keys</code> and whose value is
-        <var>keys</var>.
-        </li>
-        <li>Return <var>map</var>.
-        </li>
+        <li>Return <var>json</var>.
       </ol>
+
+      <p>
+        A <dfn>PushSubscriptionJSON</dfn> dictionary represents the <a>JSON type</a> of a
```
domenic, line 1040: Hmm I wonder why "JSON type" is showing up as code font in the output. It shouldn't, ideally.

index.html
domenic: In this case it seems like there's actually a serialization process going on here, which is left up to the UA in some way. I.e. "The user agent MUST use a serialization method that does not contain input-dependent branchs (that is, one that is constant time)". I'm not sure what's really going on here.

domenic (commented):
Agreed mostly looks good, although I'm not sure about this underlying value vs. serialized value business.

index.html
beverloo: Also done. We don't depend on the infra spec yet so migrating to that would be a follow-up.

beverloo:
Thank you for the feedback!

I guess it's a bit of both, _underlying value_ refers to the attribute's behaviour so for the dictionary we can use it. Not having branches is significant here from a timing point of view, as these operations expose cryptographic data to the developer.

index.html
beverloo: It's a clarification in either case, happy to remove it?

tobie:
Yeah, from my perspective, the serialization happens in `JSON.stringify`, not when calling `obj.toJSON()` (which just returns a JS object). But others might see things differently.

index.html
beverloo: I've left this as a comment for now and filed an issue on ReSpec:
https://github.com/w3c/respec/issues/1297

beverloo:
Cheers again, all applied. Two follow-ups: the ReSpec issue (https://github.com/w3c/respec/issues/1297) and deciding on adding a dependency on _infra_. I moved the constant time statement to the attribute.

I'll submit this, happy to address any other follow-ups :).

PR #268, opened 2017-06-22
Add support for pr-preview

tobie:
See https://github.com/tobie/pr-preview#pr-preview for more details.

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,4 @@
+{
+    "src_file": "index.html",
+    "type": "respec"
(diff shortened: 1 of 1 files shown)


beverloo:
Awesome, thanks! I see that you already enabled the integration on the repository.

tobie:
Yup. Updated the README to reflect that. Hope this will be as useful to you as it is to me. (Note you can keyboard navigate the diff view, which is really helpful. Instructions show up when hovering over the bottom right controls).

PR #274, opened 2017-06-26
Clarify requirements for push endpoint determinism and predictability

beverloo:
/cc @magnus-git

***
[Preview](https://s3.amazonaws.com/pr-preview/beverloo/push-api/security-review-feedback.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/push-api/7bc329f...beverloo:9c08033.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -96,16 +96,6 @@
               status: "Living Standard",
               publisher: "WHATWG"
             },
-            "WEBPUSH-PROTOCOL": {
-              title: "The Web Push Protocol",
-              href: "https://tools.ietf.org/html/draft-ietf-webpush-protocol",
-              authors: [
-                "Martin Thomson",
-                "Brian Raymor",
-              ],
-              status: "Internet-Draft",
-              publisher: "IETF"
-            },
             "WEBPUSH-ENCRYPTION": {
               title: "Message Encryption for Web Push",
               href: "https://tools.ietf.org/html/draft-ietf-webpush-encryption",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -526,6 +526,13 @@
         subscription</a> MUST be <a>deactivated</a>.
       </p>
       <p>
+        The <a>push endpoint</a> MUST NOT enable information about the user to be derived by actors
```
martinthomson: s/enable/expose

index.html
```diff
@@ -526,6 +526,13 @@
...
+        other than the <a>push service</a>, such as the user's device, identity or location.
+        <a>Push services</a> that do not require <a>push subscriptions</a> to be restricted to an
+        <a>application server</a> [[!WEBPUSH-VAPID]] MUST NOT generate predictable
```
martinthomson: Why is this requirement only levied on endpoints that are NOT restricted?  RFC 8030 is pretty clear on this point and restriction to an application server doesn't change that.

martinthomson (changes requested):
You should reference the text in RFC 8030 about this: https://tools.ietf.org/html/rfc8030#section-8.2

index.html

index.html
beverloo: Good point - I've updated this to refer 8030 instead.

beverloo (commented):
Thanks! Please take another look, I've applied the changes and s/WEBPUSH-PROTOCOL/RFC8030/ everywhere. I'll run tidy after the current series of changes landed.

magnus-git:
This update looks good to me, thanks!

PR #284, opened 2017-08-21
Add [SecureContext] annotations to our interfaces

### LINKED ISSUE #282: Secure contexts
Any reason this is not using `[SecureContext]`? Would be good to add a note explaining why.

beverloo:
Fixes #282

***
[Preview](https://s3.amazonaws.com/pr-preview/beverloo/push-api/secure-context.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/push-api/7bc329f...beverloo:16dca29.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -665,6 +665,7 @@ navigator.serviceWorker.register('serviceworker.js').then(
         [[!SERVICE-WORKERS]], which this specification extends.
       </p>
       <pre class="idl">
+        [SecureContext]
         partial interface ServiceWorkerRegistration {
           readonly attribute PushManager pushManager;
         };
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -665,6 +665,7 @@
         [[!SERVICE-WORKERS]], which this specification extends.
       </p>
       <pre class="idl">
+        [SecureContext]
```
martinthomson, line 668: I'm really confused by the specification of this:

> Any extended attribute specified on a partial interface definition is considered to appear on the interface itself.

But then the text later says that the attributes only apply to the partial interface.  In this case, `ServiceWorkerRegistration` is annotated in the same way, so this is OK (it's not in the notifications API though).

martinthomson (approved):

Issue #273, opened 2017-06-26
Security review of Push API W3C Working Draft 23 June 2017


magnus-git:
I am a member of the IETF Security Directorate and have been asked to review this document from a security perspective by Sam Weiler, as part of a W3C trial attempting to get broader security reviews of W3C specs before they are published as recommendations. These comments were written primarily for the benefit of said trial. Document editors and WG chairs should treat these comments just like any other comments.

Observations:
- The Security Considerations sections note that there's no way for the user agent to validate that a push message was sent by an application server having the same origin as the webapp since "the application server is able to share the details necessary to use a push subscription with a third party at its own discretion." Is this undesirable behavior? If so, I am a little puzzled by the statement since I thought the webapp's ECDH public key could be used for identification (e.g., if provided in the form of a certificate). I may miss some context here though.
- The text talks about no reuse of push endpoints for new push subscriptions etc., but seems to be silent on good practices for selecting good endpoints. For example, if a user agents selects them in a predictable manner, this would also seem to allow for situations of "cross-subscription" push messages? If that's the case, then maybe provide guidance on how push endpoints shall be generated?

Editorial:
- In the Security Considerations section, "necessary to to use" -> "necessary to use"

/M

beverloo:
Hi Magnus, thank you for the review!

> The Security Considerations sections note that there's no way for the user agent to validate that a push message was sent by an application server having the same origin as the webapp since "the application server is able to share the details necessary to use a push subscription with a third party at its own discretion." Is this undesirable behavior?

No, that's intended behaviour. Part of this is theoretical: there's no way for us to verify the originating origin or a push message if the application server decides to share their private key with other parties. Part of this is practical: there are valid use-cases for apps to do this, for instance a news feed aggregation client that receives updates from multiple agencies.

> The text talks about no reuse of push endpoints for new push subscriptions etc., but seems to be silent on good practices for selecting good endpoints.

Good point! I've uploaded a PR to address this, and would appreciate feedback: https://github.com/w3c/push-api/pull/274

> In the Security Considerations section, "necessary to to use" -> "necessary to use"

Also addressed in the PR.

magnus-git:
Thanks Peter.
On #274 , would it make sense to point to some existing resource providing guidance for how to generate such identifier? Perhaps just referring to the Privacy Considerations in RFC 8030 is sufficient, in addition to the text you suggested?
[...]

PR #286, opened 2017-08-21
Clarify that a Service Worker may be woken up and consume resources

beverloo:
Part of #273

***
[Preview](https://s3.amazonaws.com/pr-preview/beverloo/push-api/security-review-background.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/push-api/7bc329f...beverloo:eaf2e9e.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -517,6 +517,12 @@
         revocable.
       </p>
       <p>
+        The Push API may have to wake up the Service Worker associated with the <a>service worker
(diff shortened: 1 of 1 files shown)


beverloo:
We can't really require this, but this clarifies that it may happen. WDYT?

I deliberately didn't use RFC 2119 MAY in this block as it's not normatively significant.

magnus-git:
I agree this is a reasonable update. Thanks,

PR #348, opened 2022-06-01
check and consume user activation

### LINKED ISSUE #346: Does PushManager subscribe() need transient activation check?
https://w3c.github.io/push-api/#dom-pushmanager-subscribe does not mention whether a https://html.spec.whatwg.org/multipage/interaction.html#transient-activation check should be done.
AIUI, Chrome and Firefox enforce these checks and WebKit is about to add this check at https://bugs.webkit.org/attachment.cgi?id=455337&action=review.
Would it make sense to update the spec accordingly?

marcoscaceres:
Closes #346 

 * [ ] [Modified Web platform tests ](https://github.com/web-platform-tests/wpt/pull/34273)

Implementation commitment:
 
 * [x] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1772051)
 * [x] WebKit - https://bugs.webkit.org/attachment.cgi?id=455337&action=review

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/348.html" title="Last updated on Dec 8, 2025, 9:11 AM UTC (4f2f8ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/348/518d7d6...4f2f8ac.html" title="Last updated on Dec 8, 2025, 9:11 AM UTC (4f2f8ac)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1200,6 +1200,21 @@
             </li>
           </ol>
         </li>
+        <li>If |global| is a {{Window}} object:
+          <ol>
+            <li>Let |permissionState| be "`push`"'s [=permission state=].
+            </li>
+            <li>If |permissionState| is "`prompt`":
+              <ol>
+                <li>If |global| does not have [=transient activation=], return [=a promise rejected
+                with=] with a {{"NotAllowedError"}} {{DOMException}}.
+                </li>
+                <li>[=Consume user activation=] of |global|.
+                </li>
+              </ol>
+            </li>
+          </ol>
+        </li>
         <li>Run these steps [=in parallel=]:
           <ol>
             <li>Let |subscription| be null.
--- a/index.html
+++ b/index.html
@@ -1206,8 +1206,9 @@
             </li>
             <li>If |permissionState| is "`prompt`":
               <ol>
-                <li>If |global| does not have [=transient activation=], return [=a promise rejected
-                with=] with a {{"NotAllowedError"}} {{DOMException}}.
+                <li>If |global| does not have [=transient activation=], then [=queue a global
+                task=] on the [=networking task source=] using |global| to [=reject=] |promise|
+                with a {{"NotAllowedError"}} {{DOMException}} and terminate these steps.
                 </li>
                 <li>[=Consume user activation=] of |global|.
                 </li>


martinthomson (approved):

marcoscaceres:
Preparing a WPTest...

beverloo:
I'm not sure this is Web compatible -

1. `subscribe()` can be called from within a service worker, where there are no local activations that can be consumed -- particularly the [consume user activation](https://html.spec.whatwg.org/multipage/interaction.html#consume-user-activation) steps assume that there's a window.
2. Looking at our metrics, only about a fourth of permission requests happen with a gesture. We mitigate against this by displaying a different user interface (if any at all) in those cases. In my reading it would be valid for an implementation not to show any prompt at all.

saschanaz:
Note that Gecko opted to require user activation only if the permission is not granted, and just proceeds without activation if it's granted, which I think is the best way as reportedly some websites repeatedly calls `pushManager.subscribe()` every time: https://bugzilla.mozilla.org/show_bug.cgi?id=1486010#c5

>Instead, the page calls pushManager.subscribe every time you visit it

marcoscaceres:
I'll take a closer look at what WebKit does... we might be doing something similar @saschanaz.

index.html
```diff
@@ -1200,6 +1200,21 @@ <h3>
             </li>
           </ol>
         </li>
+        <li>If |global| is a {{Window}} object:
+          <ol>
+            <li>Let |permissionState| be "`push`"'s [=permission state=].
+            </li>
+            <li>If |permissionState| is "`prompt`":
+              <ol>
+                <li>If |global| does not have [=transient activation=], return [=a promise rejected
```
saschanaz: The Gecko implementation is more like saving the transient activation state and use that info later in async, but maybe that's a bit unusual to specify?

saschanaz (approved):
I'm fine with this.

index.html
annevk: What happens if someone else consumes it meanwhile?

saschanaz: Ah yes, we only check the activation but doesn't really consume it, that's why we can do that. Certainly things would change when we consume it.

annevk: I should queue a task here and not just return a fresh promise. I do think we want to consume immediately and not from a task. Otherwise it's not very predictable for web developers.

PR #350, opened 2022-06-02
Rewrite .subscribe() to fix various bugs

marcoscaceres:
The following are not testable/observable, but spec bugs nonetheless:  

  * Fix `PushSubscriptionOptionsInit` usage.
  * Do less work in parallel.
  * Handle `userVisibleOnly` member.
  * Resolve promises using networking tasks source.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/350.html" title="Last updated on Jun 30, 2022, 5:53 AM UTC (10e0d68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/350/5429a60...10e0d68.html" title="Last updated on Jun 30, 2022, 5:53 AM UTC (10e0d68)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,6 +26,12 @@
               w3cid: "68503"
             },
             {
+              name: "Marcos Caceres",
+              company: "Apple Inc.",
+              companyURL: "https://www.apple.com/",
+              w3cid: "39125"
+            },
+            {
               name: "Bryan Sullivan",
               company: "AT&T",
               companyURL: "http://www.att.com/",
@@ -569,20 +575,30 @@
         The <dfn>subscribe()</dfn> method when invoked MUST run the following steps:
       </p>
       <ol class="algorithm">
-        <li>If the |options| argument has a {{PushSubscriptionOptionsInit/userVisibleOnly}} value
-        set to `false` and the user agent requires it to be `true`, return [=a promise rejected
-        with=] an {{"NotAllowedError"}} {{DOMException}}.
-        </li>
-        <li>If the |options| argument does not include a non-null value for the
-        {{PushSubscriptionOptionsInit/applicationServerKey}} member, and the <a>push service</a>
-        requires one to be given, return [=a promise rejected with=] a {{"NotSupportedError"}}
-        {{DOMException}}.
-        </li>
         <li>Let |promise| be [=a new promise=].
         </li>
         <li>Let |global| be [=this=]' [=relevant global object=].
         </li>
         <li>Return |promise| and continue [=in parallel=].
+          <aside class="note" title="Validation order can vary across user agents">
+            <p>
+              Because of implementation-specific reasons, user agents are known to do some of the
+              following checks in different order (e.g., some check if
+              {{PushSubscriptionOptionsInit/userVisibleOnly}} is allowed after validating the
+              {{PushSubscriptionOptionsInit/applicationServerKey}}, and vice versa). However, we
+              don't believe this affects interoperability of implementations or web applications.
+            </p>
+          </aside>
+        </li>
+        <li>If the |options| argument has a {{PushSubscriptionOptionsInit/userVisibleOnly}} value
+        set to `false` and the user agent requires it to be `true`, [=queue a global task=] on the
+        [=networking task source=] using |global| to [=reject=] |promise| {{"NotAllowedError"}}
+        {{DOMException}}
+        </li>
+        <li>If the |options| argument does not include a non-null value for the
+        {{PushSubscriptionOptionsInit/applicationServerKey}} member, and the <a>push service</a>
+        requires one to be given, [=queue a global task=] on the [=networking task source=] using
+        |global| to [=reject=] |promise| with a {{"NotSupportedError"}} {{DOMException}}.
         </li>
         <li>If the |options| argument includes a non-null value for the
         {{PushSubscriptionOptions/applicationServerKey}} attribute, run the following sub-steps:
@@ -593,14 +609,14 @@
             {{PushSubscriptionOptionsInit/applicationServerKey}} using the base64url encoding
             [[RFC7515]].
             </li>
-            <li>If decoding fails, [=queue a global task=] on the [=user interaction task source=]
-            using |global| to [=reject=] |promise| with an {{"InvalidCharacterError"}}
-            {{DOMException}} and terminate these steps.
+            <li>If decoding fails, [=queue a global task=] on the [=networking task source=] using
+            |global| to [=reject=] |promise| with an {{"InvalidCharacterError"}} {{DOMException}}
+            and terminate these steps.
             </li>
             <li>Ensure that |options|'s {{PushSubscriptionOptionsInit/applicationServerKey}}
             describes a valid point on the P-256 curve. If its value is invalid, [=queue a global
-            task=] on the [=user interaction task source=] using |global| to [=reject=] |promise|
-            with an {{"InvalidAccessError"}} {{DOMException}} and terminate these steps.
+            task=] on the [=networking task source=] using |global| to [=reject=] |promise| with an
+            {{"InvalidAccessError"}} {{DOMException}} and terminate these steps.
             </li>
           </ol>
         </li>
@@ -608,8 +624,8 @@
         registration</a>.
         </li>
         <li>If |registration|'s [=service worker registration/active worker=] is null, [=queue a
-        global task=] on the [=user interaction task source=] using |global| to [=reject=]
-        |promise| with an {{"InvalidStateError"}} {{DOMException}} and terminate these steps.
+        global task=] on the [=networking task source=] using |global| to [=reject=] |promise| with
+        an {{"InvalidStateError"}} {{DOMException}} and terminate these steps.
         </li>
         <li>Let |sw| be |registration|'s [=service worker registration/active worker=].
         </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -195,13 +195,16 @@ <h2>
           subscription</a> having the new keys as |newSubscription|.
         </p>
         <p>
-          To <dfn>create a push subscription</dfn>, given an <a>PushSubscriptionOptions</a> object
-          of |options|, the <a>user agent</a> must run the following steps:
+          To <dfn>create a push subscription</dfn>, given an {{PushSubscriptionOptionsInit}}
+          |optionsDictionary|:
```
marcoscaceres:
```suggestion
          |optionsDictionary:PushSubscriptionOptionsInit|:
```

index.html
```diff
@@ -195,13 +195,16 @@ <h2>
...
-          of |options|, the <a>user agent</a> must run the following steps:
+          To <dfn>create a push subscription</dfn>, given an {{PushSubscriptionOptionsInit}}
+          |optionsDictionary:PushSubscriptionOptionsInit|:
         </p>
-        <ol>
+        <ol class="algorithm">
           <li>Let |subscription| be a new <a>push subscription</a>.
           </li>
-          <li>Set the `options` attribute of |subscription| to be a copy of |options|.
+          <li>Let |options| be a newly created {{PushSubscriptionOptions}} object, initializing its
```
marcoscaceres:
```suggestion
          <li>Let |options: PushSubscriptionOptions| be a newly created {{PushSubscriptionOptions}} object, initializing its
```

index.html
```diff
@@ -195,13 +195,16 @@ <h2>
...
-          of |options|, the <a>user agent</a> must run the following steps:
+          To <dfn>create a push subscription</dfn>, given an {{PushSubscriptionOptionsInit}}
+          |optionsDictionary:PushSubscriptionOptionsInit|:
         </p>
-        <ol>
+        <ol class="algorithm">
           <li>Let |subscription| be a new <a>push subscription</a>.
           </li>
-          <li>Set the `options` attribute of |subscription| to be a copy of |options|.
+          <li>Let |options: PushSubscriptionOptions| be a newly created {{PushSubscriptionOptions}} object, initializing its
```
marcoscaceres:
```suggestion
          <li>Let |options:PushSubscriptionOptions| be a newly created {{PushSubscriptionOptions}} object, initializing its
```

index.html
```diff
@@ -195,13 +195,16 @@ <h2>
...
+          To <dfn>create a push subscription</dfn>, given an {{PushSubscriptionOptionsInit}}
+          |optionsDictionary:PushSubscriptionOptionsInit|:
         </p>
-        <ol>
+        <ol class="algorithm">
           <li>Let |subscription| be a new <a>push subscription</a>.
           </li>
-          <li>Set the `options` attribute of |subscription| to be a copy of |options|.
+          <li>Let |options:PushSubscriptionOptions| be a newly created {{PushSubscriptionOptions}}
+          object, initializing its attribute with the corresponding values of |optionsDictionary|.
```
marcoscaceres:
```suggestion
          object, initializing its attributes with the corresponding values of |optionsDictionary|.
```

marcoscaceres:
```suggestion
          object, initializing its attributes with the corresponding values of |optionsDictionary|.
```

index.html
```diff
@@ -548,73 +551,92 @@ <h2>
...
       <p>
-        The <dfn>subscribe</dfn> method when invoked MUST run the following steps:
+        The <dfn>subscribe()</dfn> method when invoked MUST run the following steps:
       </p>
-      <ol>
-        <li>Let |promise| be <a>a new promise</a>.
+      <ol class="algorithm">
+        <li>If the user agent requires |options|'s {{PushSubscriptionOptionsInit/userVisibleOnly}}
+        member to be `true`, return [=a promise rejected with=] an {{"NotAllowedError"}}
+        {{DOMException}}.
```
martinthomson: This would always reject if userVisibleOnly had to be true, rather than just when it is false.

```suggestion
        <li>If the |options| argument has a {{PushSubscriptionOptionsInit/userVisibleOnly}} value set to `false` and 
        the user agent requires it to be `true`, return [=a promise rejected with=] an {{"NotAllowedError"}}
        {{DOMException}}.
```

I'm not 100% on the bit where we this leans on defaults, your next step seems to do that though.

index.html
```diff
@@ -548,73 +551,92 @@ <h2>
...
+            </li>
+            <li>If decoding fails, return [=a promise rejected with=] a {{"InvalidCharacterError"}}
+            {{DOMException}}.
             </li>
-            <li>Ensure that |applicationServerKey| describes a valid point on the P-256 curve. If
-            the |applicationServerKey| value is invalid, reject |promise| with a {{DOMException}}
-            whose name is {{"InvalidAccessError"}} and terminate these steps.
+            <li>Ensure that |options|'s {{PushSubscriptionOptionsInit/applicationServerKey}}
+            describes a valid point on the P-256 curve. If its value is invalid, return [=a promise
+            rejected with=] an {{"InvalidAccessError"}} {{DOMException}}.
```
martinthomson: In terms of synchronous steps, this is not one that can be done synchronously in all implementations as it is non-trivial in the sense that it invokes cryptography.  It is possible that this difference would be observable.

(I checked the gecko implementation and I can't see where we implement this check, so season accordingly.  We probably implement this check on the push service instead, which we should probably fix, but there you go...)

index.html
```diff
@@ -548,73 +551,92 @@ <h2>
...
+        <li>If |sw| is already subscribed, run the following substeps:
           <ol>
-            <li>Retrieve the <a>push subscription</a> associated with the <a>Service Worker</a>.
+            <li>Let |subscription| be the <a>push subscription</a> associated with |sw|.
             </li>
-            <li>If there is an error, reject |promise| with a {{DOMException}} whose name is
-            {{"AbortError"}} and terminate these steps.
+            <li>If |subscription| is an error, [=queue a global task=] on the [=networking task
+            source=] using |global| to [=reject=] |promise| with an {{"AbortError"}}
+            {{DOMException}} and terminate these steps.
```
martinthomson: How can *subscription* be an error here?  The outcome of an attempt to retrieve it might return an error, but the *subscription* will be a subscription, right?

Maybe the answer is that you need to invoke a process for obtaining *subscription* from *sw*, which could then return an error.  In that case, you won't learn that *sw* is already subscribed, so you should rearrange these steps along the lines of:

1. Let *subscription* be the result of obtaining the current push subscription from *sw*.
2. If *subscription* is an error, (...this step...reject the promise)
3. If *subscription* exists, run the following substeps:
    a. (...the other checks here...)
    b. (...resolve the promise.)
4. (Otherwise) (...make a subscription)

index.html
marcoscaceres: Yeah, I was wondering about this also. I'll switch it back to being done async.

index.html
marcoscaceres: > How can subscription be an error here? 

You are right. It can't be an error. 

> The outcome of an attempt to retrieve it might return an error, but the subscription will be a subscription, right?

Right. In fact, so long as [=create a push subscription=] succeeded previously, there is _always_ a subscription. 

The problem (question?) is actually in [=create a push subscription=]: can [=create a push subscription=] result in an exception? 

It appears it can for a variety of reasons (e.g., [Gecko binding layer](https://searchfox.org/mozilla-central/source/__GENERATED__/dom/bindings/PushManagerBinding.cpp#1637) generates a bunch, [WebKit](https://webkit-search.igalia.com/webkit/source/Source/WebCore/workers/service/ServiceWorkerContainer.cpp#573) re-throws by rejecting).

So, I think the right thing to do here is basically say, if creating [=create a push subscription=] results in an exception, reject the promise with it. 

I'll also update [=create a push subscription=] so it "tries" to create the push subscription and re-throws any exceptions.

marcoscaceres: Sorry, to also clarify, you are right that the retrieval of the conceptual "subscription" might result in an error (basically, what happens with `getSubscription()`).

index.html
```diff
@@ -548,73 +562,94 @@ <h2>
...
         </li>
-        <li>Return |promise| and continue the following steps asynchronously.
+        <li>If the |options| argument does not include a non-null value for the
+        {{PushSubscriptionOptionsInit/applicationServerKey}} member, and the <a>push service</a>
+        requires one to be given, return [=a promise rejected with=] a {{"NotSupportedError"}}
+        {{DOMException}}.
+        </li>
+        <li>Let |promise| be [=a new promise=].
+        </li>
+        <li>Let |global:Window| be [=this=]' [=relevant global object=].
```
beverloo: The relevant global object within a service worker isn't a window, so this definition may be invalid. None of the uses of `global` seem to require it to be a Window though, the worker global (in my reading) also has a task queue.

beverloo (approved):
Generally these changes look good to me, modulo one type hint that should be generalized.

Our implementation runs one of the steps out of order: the `userVisibleOnly` requirement is only checked quite a late in the process, as there may be (toady hypothetical) cases in which we don't require it, e.g. for extensions.

index.html
marcoscaceres: Ah, yeah. good point. I'll remove the type.

marcoscaceres:
> Our implementation runs one of the steps out of order: the userVisibleOnly requirement is only checked quite a late in the process, as there may be (toady hypothetical) cases in which we don't require it, e.g. for extensions.

Ok, I'll move that to async part and just add a note. The ordering when that happens doesn't matter too much.

Issue #360, opened 2023-09-11
Declarative Web Push


beidson:
### Declarative Web Push:
More efficient and privacy-preserving push notifications

By: Brady Eidson
Technical review by: Marcos Caceres and Anne van Kesteren. 

This explainer proposes various changes to existing web standards so that, in the majority of cases, push-initiated notifications can be presented by the platform or user agent without involving a Service Worker.

The primary mechanism to enable this is by making the push message payload a declarative description of the notification to be displayed.

Our proposed model doesn’t require a Service Worker work to get a PushSubscription. Even if a Service Worker is registered, push notification messages bypass it by default.
This makes push notifications from the web more privacy-preserving as sites aren’t given the opportunity to execute any JavaScript until a user explicitly interacts with a notification. It also makes web push notifications more efficient by skipping the CPU and battery cost required to launch a Service Worker and execute its code.

We also propose a more flexible event handling model that developers can opt into to transform a push notification. This new model still depends on Service Workers, but provides the privacy-preserving guarantee that there is always a fallback notification to be shown, so pushes cannot trigger silent background runtime.

This explainer describes what we have implemented internally to prototype or what we plan to implement next.
Implementation experience and feedback from the standards community might change some details.

### What are the drawbacks of current push notifications?

The W3C’s Push API  and related specifications work great for maximizing user engagement, but suffer from notable drawbacks: 

* Requires a Service Worker which has higher web developer complexity, and which cost significant CPU, battery, and potentially network resources to present a notification.
* Because a Service Worker is required, user agents with certain types of privacy features (such as Safari’s Intelligent Tracking Prevention) need to make hard trade-offs between push reliability and effective privacy.
* User agents require a fallback plan for when the web app doesn’t fulfill their promise of displaying a user visible notification. 

### Learning from platform experience

Push notifications that come in to an OS platform are usually displayed by the platform itself, without any app code executing.
It is also fairly standard practice for apps to optionally get a small amount of CPU time to transform an incoming push notification - specially where changes since the last push message matter. Some examples of tasks performed during these allocated time slices include decrypting push data, updating the badge count, or updating based on score from an in-progress game. 
If the app fails to complete a transformation task in their allotted time - due to coding error or other factors outside their control - then the original notification is displayed to a user. This automatic recovery model avoids disrupting the user experience, and ensures that push notifications don’t become a vector for untrusted silent background runtime.

We have deep experience with this flow and it is the base model for our proposed enhancements to Web Push.
[...]

PR #368, opened 2023-12-14
Expose pushManager on Navigator

marcoscaceres:
Part of https://github.com/w3c/push-api/issues/360

This PR introduces: 

 * navigator.pushManager
 * makes changes to the various method/algorithms of `pushManager` to distinguish between service worker scope and running in a regular window.

Discussion (or stuff I'm not super happy with):

* Using the "environment settings object" is probably not the best here, because Workers are also an environment (or have an environment settings object).

The current spec is currently hand-wavy about how push registrations are persistently associated with a service worker (which is ok), but not it gets a little bit messier because we need to distinguish between the Window object associated registration and the service worker ones. 

Anyone got any suggestions as to what I could use to hand the registration off instead?     

 

Implementation commitment:
 
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)
 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/368.html" title="Last updated on Dec 14, 2023, 5:08 AM UTC (95aeb98)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/368/d8401e8...95aeb98.html" title="Last updated on Dec 14, 2023, 5:08 AM UTC (95aeb98)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -63,9 +63,9 @@
         a <a>push service</a>. An <a>application server</a> can send a <a>push message</a> at any
         time, even when a web application or <a>user agent</a> is inactive. The <a>push service</a>
         ensures reliable and efficient delivery to the <a>user agent</a>. <a>Push messages</a> are
-        delivered to a <a>Service Worker</a> that runs in the origin of the web application, which
-        can use the information in the message to update local state or display a notification to
-        the user.
+        delivered to a [=push subscription owner=] that runs in the origin of the web application,
+        which can use the information in the message to update local state or display a
+        notification to the user.
       </p>
       <p>
         This specification is designed for use with the <a>web push protocol</a>, which describes
@@ -172,8 +172,15 @@
         <p>
           A <dfn>push subscription</dfn> is a message delivery context established between the
           <a>user agent</a> and the <a>push service</a> on behalf of a web application. Each
-          <a>push subscription</a> is associated with a <a>service worker registration</a> and a
+          <a>push subscription</a> is associated with a [=push subscription owner=].
+        </p>
+        <p>
+          A <dfn>push subscription owner</dfn> refers to a <a>service worker registration</a>. A
           <a>service worker registration</a> has at most one <a>push subscription</a>.
+        </p>
+        <p class="note">
+          Future updates to this specification could expand the definition of a [=push subscription
+          owner=] to include other types, broadening the range of possible implementations.
         </p>
         <p>
           A <a>push subscription</a> has an associated <dfn>push endpoint</dfn>. It MUST be the
@@ -431,13 +438,13 @@
         <li>the <a>push service</a> delivers the message to a specific <a>user agent</a>,
         identifying the <a>push endpoint</a> in the message;
         </li>
-        <li>the <a>user agent</a> identifies the intended <a>Service Worker</a> and activates it as
-        necessary, and delivers the <a>push message</a> to the <a>Service Worker</a>.
+        <li>the <a>user agent</a> identifies the intended [=push subscription owner=] and activates
+        it as necessary, and delivers the <a>push message</a> to the [=push subscription owner=].
         </li>
       </ul>
       <p>
-        This overall framework allows <a>application servers</a> to activate a <a>Service
-        Worker</a> in response to events at the <a>application server</a>. Information about those
+        This overall framework allows <a>application servers</a> to activate a [=push subscription
+        owner=] in response to events at the <a>application server</a>. Information about those
         events can be included in the <a>push message</a>, which allows the web application to
         react appropriately to those events, potentially without needing to initiate network
         requests.
--- a/index.html
+++ b/index.html
@@ -175,12 +175,29 @@
           <a>push subscription</a> is associated with a [=push subscription owner=].
         </p>
         <p>
-          A <dfn>push subscription owner</dfn> refers to a <a>service worker registration</a>. A
-          <a>service worker registration</a> has at most one <a>push subscription</a>.
+          A <dfn>push subscription owner</dfn> can be one of the following:
         </p>
+        <dl>
+          <dt>
+            [=Service worker registration=]
+          </dt>
+          <dd>
+            A [=service worker registration=] has at most one [=push subscription=] associated with
+            it.
+          </dd>
+          <dt>
+            [=Environment settings object=]
+          </dt>
+          <dd>
+            This represents the security origin for the [=push subscription=]. An [=environment
+            settings object=] has at most one [=push subscription=] associated with it.
+          </dd>
+        </dl>
         <p class="note">
-          Future updates to this specification could expand the definition of a [=push subscription
-          owner=] to include other types, broadening the range of possible implementations.
+          This specification currently details handling of push subscriptions in the context of
+          both [=service worker registrations=] and an [=environment settings object=]. Future
+          updates could further expand the definition of a [=push subscription owner=] to encompass
+          additional types, further broadening the range of possible implementations.
         </p>
         <p>
           A <a>push subscription</a> has an associated <dfn>push endpoint</dfn>. It MUST be the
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Actually, thinking about this... if an (main thread) origin (i.e., "a site") has a persistent environment settings object, and service worker has its own independently persistent environment-settings object, then we could say that every [environment](https://html.spec.whatwg.org/multipage/webappapis.html#environment) has an associated push subscription (which can be `null`).
 
This could work because: 

 * a service worker can has at most 1 push subscription
 * a window would have at most 1 push subscription. 
 
Tying subscription to the environment then that could answer how push subscriptions are associated with execution contexts.

martinthomson:
Thanks for writing this up Marcos.  I'm going to probably step aside on this one and defer to @saschanaz.  Both of us are about to go on some leave, so I hope you don't mind if this takes a little while.

marcoscaceres:
No problem at all. I'll put up a few PRs over xmas and try to keep them small. 

I talked to Anne about using "environment", but he suggested they are too volatile... so, I'm going to create some kind of persistent "push subscriptions store" instead. That will be just simple store that handles the association between push subscriptions, service workers, and the origin.

index.html
```diff
@@ -620,33 +658,30 @@ <h3>
...
-        registration</a>.
+        <li>Let |environment| be the [=relevant settings object=] for |global|.
         </li>
-        <li>If |registration|'s [=service worker registration/active worker=] is null, [=queue a
-        global task=] on the [=networking task source=] using |global| to [=reject=] |promise| with
-        an {{"InvalidStateError"}} {{DOMException}} and terminate these steps.
+        <li>Let |sw| be the |environment|'s [=environment/active service worker=].
         </li>
-        <li>Let |sw| be |registration|'s [=service worker registration/active worker=].
+        <li>If |sw| is null and |global| is a {{ServiceWorkerRegistration}}, [=queue a global
```
saschanaz, line 665: ServiceWorkerGlobalScope?

saschanaz:
Some early thought:

> I talked to Anne about using "environment", but he suggested they are too volatile... so, I'm going to create some kind of persistent "push subscriptions store" instead. That will be just simple store that handles the association between push subscriptions, service workers, and the origin.

Perhaps we can associate push manager to storage bottle as Web Locks does for lock managers?: https://w3c.github.io/web-locks/#lock-managers

annevk:
I was looking at tackling this but ran into #384. Because #360 actually describes something that I think we can use. It suggests that a "window push subscription" is 1:1 with a push subscription for a service worker registration whose scope is `/` (on the same origin).

I think we should just create a service worker registration for that scope when none exists and bypass the active worker check (which seems to be there mainly for consistency).

An issue that we run into, but the existing specification also runs into, is this line in the Service Workers specification:

> A [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) of an identical [scope url](https://www.w3.org/TR/service-workers/#dfn-scope-url) when one already exists in the user agent causes the existing [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) to be replaced.

But it seems to me it could just as well say that it's being updated in place (or that some state, such as the push subscription, is carried over).

This setup has these benefits:

1. Clarifies the lifetime by tying it to service worker registration lifetime. No need to integrate with Storage (though Service Workers needs to of course).
2. Creates interoperability with existing push subscriptions on scope `/`.
3. It is very simple. No need for more complicated ownership structures or anything. Only fairly modest changes to the `PushManager` class.

Thoughts?

annevk:
@jakearchibald I know it's been a while, but maybe you could have a look at https://github.com/w3c/push-api/pull/368#issuecomment-2233366143 given your experience with service workers. I'd appreciate it!

mkruisselbrink:
> An issue that we run into, but the existing specification also runs into, is this line in the Service Workers specification:
> 
> > A [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) of an identical [scope url](https://www.w3.org/TR/service-workers/#dfn-scope-url) when one already exists in the user agent causes the existing [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) to be replaced.
> 
> But it seems to me it could just as well say that it's being updated in place (or that some state, such as the push subscription, is carried over).

In practice I don't think that sentence in the service worker spec means much. The actual algorithms around registering service workers (i.e. starting with https://w3c.github.io/ServiceWorker/#start-register-algorithm) already update an existing registration in place rather than creating a new one if the scope url matches.

If I understand it correctly, I believe you're proposing introducing a service worker registration that doesn't have any service workers associated with it. That does seem a bit odd to me, although I don't think anything would necessarily break by changing that assumption... I think there might be some subtlety about what should happen when a service worker is later registered at that same scope but installation fails (should that unregister the "/" registration as currently spec'ed, or would we want to go back to having a registration without a service worker?).

asutherland:
Having a Service Worker Registration without an active ServiceWorker feels like it discards a fairly fundamental invariant that may currently be reflected quite deeply into existing browser implementations.  (It would not be a minor thing to relax this invariant in Gecko.)

Can push subscription storage just be a map from (WindowClient or ServiceWorkerRegistration) to (PushSubscription)?

And then we can sprinkle some "Eligible PushSubscription" dust on top for lifetime management of the contents of the map to deal with the fact that browsers have complex session management things going on, especially on resource-constrained devices and where the ServiceWorkers spec has not concretely addressed https://github.com/w3c/ServiceWorker/issues/626.  This allows browsers latitude in retaining push subscriptions for WindowClients that aren't currently loaded.

Also, this avoids complications with magic treatment of the scope "/" potentially interacting weirdly with the proposal at https://github.com/w3c/ServiceWorker/issues/1512.

annevk:
Thanks both for weighing in!

1. I think the overloading is nice as it allows for reusing an existing subscription. And that ends up working in both directions. And of course you don't have to implement it as a dummy "service worker registration", that would be more of a specification device.
2. I don't think a window client is a viable key as we wouldn't want two `https://example.com/` windows to end up with separate subscriptions. In fact, `https://example.com/` and `https://example.com/bar` should share a subscription too. (A `ServiceWorkerRegistration` also doesn't seem like it would be persistent enough, but maybe a "service worker registration" is.)
3. I see your point on keying of service worker registrations potentially changing, but that also would have to be opt-in and therefore it doesn't really clash with this scheme and the default would still end up being straightforward and sharing a subscription.

annevk:
The more I think about this the more I'm inclined to stick with a service worker registration as specification device as it just answers so many questions that would otherwise have to be answered in some other way for this rather small scenario.

asutherland:
> 1. I think the overloading is nice as it allows for reusing an existing subscription. And that ends up working in both directions. And of course you don't have to implement it as a dummy "service worker registration", that would be more of a specification device.

I'm confused by this.  Are you suggesting that if a window registers a push subscription and the origin did not previously have a service worker registration for the scope "/" then a browser would be equally compliant if getRegistrations() returned or did not return a new registration with scope "/"?

annevk:
Thanks. I need to think about the best approach some more. I forgot how web-exposed registrations are. It seems annoying to have to almost duplicate them, but maybe that's the best we can do.

asutherland:
Maybe:
- the push subscriptions could be defined as operating in terms of a service worker scope
- the scope "/" can be treated specially, with it being allowed for there to be a push subscription for scope "/" without there actually being a registration.  And it would be mandated that we would not actually create a registration for "/", so getRegistrations() after a window push subscription would not change anything.
- the push subscription algorithms can use the [scope to job queue map](https://w3c.github.io/ServiceWorker/#dfn-scope-to-job-queue-map) and maybe even the job coalescing system for registering/unregistering in order to provide sequencing sanity.
  - Right now the current hand-waving of "Each [push subscription](https://w3c.github.io/push-api/#dfn-push-subscription) is associated with a [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration)" somewhat side-steps the need to be explicit about this.  But step 7 of [the subscribe method](https://w3c.github.io/push-api/#subscribe-method) "Let registration be [this](https://webidl.spec.whatwg.org/#this)'s associated [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration)." arguably already mandates this.

annevk:
@asutherland I just wanted to belatedly thank you for your suggestion there. I'm pretty close to being able to provide a new take on this PR that implements that. There might still be some hand-waving, but I think it improves upon the status quo.

I will create it as a new PR as the approach changed quite a bit, but will cross-link this one.

annevk:
Closing this in favor of #393. Review appreciated!

Issue #380, opened 2024-07-09
PushMessageData is poorly defined and seems broken


annevk:
1. Most methods refer to a _bytes_ variable that is not defined. Presumably this should be the internal slot that references the associated byte sequence or null?
2. The methods don't handle _bytes_ being null. Though perhaps it should never be null and that should instead be handled one layer up? It does seem that `data` on `PushEvent` is nullable, though `data` on `PushEventInit` is not??? That would match how section 10.4 initializes this.

annevk:
What remains here is making `data` on `PushEventInit` nullable (and default to null). That's a normative change that requires changes to implementations, but I think would make sense given the overall desired model.

annevk:
This was fixed in https://github.com/w3c/push-api/pull/385.

PR #382, opened 2024-07-09
Correct PushMessageData

### LINKED ISSUE #381: json() should be defined in terms of Infra
It should use https://infra.spec.whatwg.org/#parse-json-bytes-to-a-javascript-value directly.

annevk:
Give PushMessageData a clear internal concept to hold the bytes and also make it clear that it cannot be null. None of the methods were prepared to handle it being null and nothing ever creates a PushMessageData object where it is null.

Partially addresses the concerns in #380 and fixes #381.

---

I installed tidy-html5 but that appears to not work. There are no instructions provided by this repository as to how to do formatting so I guess I'm fine with that happening in an automatic follow-up cleanup commit...

One thing that I'm not familiar with is ReSpec and how it does linking so that probably needs careful review here as I'm trying to introduce some new links.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/382.html" title="Last updated on Jul 10, 2024, 12:13 PM UTC (cfed03d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/382/f30bf3d...cfed03d.html" title="Last updated on Jul 10, 2024, 12:13 PM UTC (cfed03d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1004,8 +1004,8 @@
         <a>parsing JSON bytes to a JavaScript value</a> given <a>this</a>'s <a for=PushMessageData>bytes</a>.
       </p>
       <p data-cite="encoding">
-        The <dfn>text</dfn> method, when invoked, MUST return the result of running <a>UTF-8
-        decode</a> on |bytes|.
+        The <dfn>text()</dfn> method steps are to return the result of running <a>UTF-8 decode</a>
+        on <a>this</a>'s <a for=PushMessageData>bytes</a>.
       </p>
       <p>
         To <dfn>extract a byte sequence</dfn> from |object|, run these steps:
--- a/index.html
+++ b/index.html
@@ -982,30 +982,30 @@
         };
       </pre>
       <p>
-        <a>PushMessageData</a> objects have an associated <dfn data-dfn-for=PushMessageData>bytes</dfn> (a
+        <a>PushMessageData</a> objects have an associated <dfn>bytes</dfn> (a
         [=byte sequence=]), which is set on creation.
       </p>
       <p>
         The <dfn>arrayBuffer()</dfn> method steps are to return an {{ArrayBuffer}} whose contents
-        are <a>this</a>'s <a data-dfn-for=PushMessageData>bytes</a>. Exceptions thrown during the creation of
+        are [=this=]'s [=bytes=]. Exceptions thrown during the creation of
         the {{ArrayBuffer}} object are re-thrown.
       </p>
       <p>
         The <dfn>blob()</dfn> method steps are to return a new {{Blob}} object whose contents are
-        <a>this</a>'s <a data-dfn-for=PushMessageData>bytes</a>.
+        [=this=]'s [=bytes=].
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -982,31 +982,30 @@ <h2>
         };
       </pre>
       <p>
-        <a>PushMessageData</a> objects have an associated [=byte sequence=] set on creation, which
-        is `null` if there was no data in the <a>push message</a>.
+        <a>PushMessageData</a> objects have an associated <dfn data-dfn-for=PushMessageData>bytes</dfn> (a
```
saschanaz: Is `data-dfn-for` needed when `<section>` already have it above?

index.html
```diff
@@ -982,31 +982,30 @@ <h2>
...
-        are re-thrown.
+        The <dfn>arrayBuffer()</dfn> method steps are to return an {{ArrayBuffer}} whose contents
+        are <a>this</a>'s <a data-dfn-for=PushMessageData>bytes</a>. Exceptions thrown during the creation of
+        the {{ArrayBuffer}} object are re-thrown.
       </p>
       <p>
-        The <dfn>blob()</dfn> method, when invoked, MUST return a {{Blob}} whose contents are
-        |bytes| and |type| is not provided.
+        The <dfn>blob()</dfn> method steps are to return a new {{Blob}} object whose contents are
+        <a>this</a>'s <a data-dfn-for=PushMessageData>bytes</a>.
```
saschanaz:
```suggestion
        [=this=]'s [=PushMessageData/bytes=].
```

I think this works... same for others.

index.html
annevk: I don't know. I could try I suppose. Is it r+ with that and the below nit?

index.html
annevk: It probably does, but that doesn't match the current editorial convention?

saschanaz: I believe that convention is just because of this file is old and we don't have to maintain high verbosity 😛

index.html
annevk: Does that apply it to both the `<dfn>` and the `<a>`s in that section? So if you reference outside you'd have to technically override it?

index.html
saschanaz: BTW is it intentional to remove the mention of type?

saschanaz (commented):
It seems PushMessageData is only set for PushEvent if any buffer exists and otherwise `data` becomes null, and PushMessageData itself does not expect null. [Gecko does not expect null either](https://searchfox.org/mozilla-central/rev/ab488b411ff15c2e11dbd4d6be0455bb64bc8fc1/dom/serviceworkers/ServiceWorkerEvents.h#221), so it matches the implementation. (I assume you checked WebKit, and my quick check says [WebKit does not either](https://searchfox.org/wubkat/rev/df1e46cebe93c020392645e94f67469843cd8063/Source/WebCore/Modules/push-api/PushMessageData.h#44).)

index.html
annevk: Yeah, "_type_ is not provided" doesn't mean anything.

saschanaz (approved):
👍

PR #385, opened 2024-08-26
Add Declarative Web Push

annevk:
This introduces a new feature whereby push messages conforming to a certain JSON format directly create an end user notification and show it (possibly preceded by an enhanced push event).

This builds on https://github.com/whatwg/notifications/pull/213 which adds URL members to notifications.

Exposing PushManager outside of service workers is #393.

---

 * [x] Modified Web platform tests (link to pull request): IDL will be tested, but otherwise there's no good infrastructure for this as-of-yet.

Implementer support:
 
 * [ ] Chromium (has indicated support at times)
 * [x] Gecko 
 * [x] WebKit

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/385.html" title="Last updated on Aug 6, 2025, 2:11 PM UTC (90707d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/385/fe3f8ff...90707d9.html" title="Last updated on Aug 6, 2025, 2:11 PM UTC (90707d9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1724,8 +1724,8 @@
                       <li>
                         <p>
                           Let |result| be the result of [=fire a push event|firing a push event=]
-                          given |registration|, null, and a new {{Notification}} object representing
-                          |notification|.
+                          given |registration|, null, and a new {{Notification}} object
+                          representing |notification|.
                         </p>
                       </li>
                       <li>
@@ -1784,8 +1784,8 @@
         </p>
         <p>
           To <dfn>fire a push event</dfn> given a [=/service worker registration=] |registration|,
-          {{PushMessageData}} object or null |data|, and a [=/notification=] or null |notification|,
-          run these steps. They return failure or a [=/push event result=].
+          {{PushMessageData}} object or null |data|, and a [=/notification=] or null
+          |notification|, run these steps. They return failure or a [=/push event result=].
         </p>
         <ol>
           <li>
@@ -1830,9 +1830,9 @@
               </li>
               <li>
                 <p>
-                  Set |notificationResult| to true if
-                  {{ServiceWorkerRegistration/showNotification()}} has been invoked; otherwise
-                  false.
+                  Set |notificationResult| to |registration|'s <a data-dfn-for=
+                  "service worker registration">has <code>showNotification()</code> been
+                  successfully invoked</a>.
                 </p>
               </li>
             </ol>
--- a/index.html
+++ b/index.html
@@ -56,7 +56,8 @@
       };
     </script>
   </head>
-  <body data-cite="service-workers FILEAPI secure-contexts hr-time permissions ECMASCRIPT">
+  <body data-cite=
+  "SERVICE-WORKERS FILEAPI SECURE-CONTEXTS HR-TIME PERMISSIONS ECMASCRIPT NOTIFICATIONS BADGING">
     <section id="abstract">
       <p>
         The <cite>Push API</cite> enables sending of a <a>push message</a> to a web application via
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1145,14 +1444,118 @@ <h2>
...
+                      <li>
+                        <p>
+                          Wait for all of the promises in the [=ExtendableEvent/extend lifetime
+                          promises=] of the dispatched event to resolve.
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          If a notification has been shown through {{showNotification()}} at this
+                          point, then abort these steps.
```
martinthomson: This seems non-ideal.  I don't know how this is supposed to work, but this requires a trap in `showNotification` to track, but you aren't monkey-patching that.  Does `preventDefault` not work?

index.html
```diff
@@ -164,6 +164,281 @@ <h2>
...
+            <li>
+              <p>
+                Let <var>appBadge</var> be null.
+              </p>
+            </li>
+            <li>
+              <p>
+                If <var>message</var>["`app_badge`"] [=map/exists=] and
+                <var>message</var>["`app_badge`"] is an integer in the range 0 to
+                9223372036854775807, inclusive, then set <var>appBadge</var> to
```
martinthomson: 2^63-1.  You could say as much and not have people break out a calculator.  Is there a constant in the badging part of the API that you can refer to instead?

Second, given that you define the value above with this range, do you need to add this check?

Finally, why?  I can't believe that any badging UX will ever show anything that large (most stop after 9 or 99 in my experience).

martinthomson (commented):
I'd like to see more explanatory text attached to this sort of change. (In the spec, not the PR, for avoidance of doubt.)

From what I can see, a message is opportunistically parsed as JSON.  If it parses and there is a "web_push": 9001 (?!) attribute, the browser attempts to make a notification.  If that works, the notification is shown.

There is also a mutable attribute attached, which would allow the SW the option of intercepting the notification and tweaking it before it is shown.  That would somewhat negate the benefits from a purely declarative notification, so it seems an unnecessary feature (the app could save the "web_push": 9001 bytes and just make a notification for itself).

index.html
annevk: It's 2^53-1, it specifies the limit in exactly the same fashion as IDL does, and the description of the format doesn't actually come with implementation requirements.

index.html
annevk: I mean, this statement is monkey patching that. Certainly seems reasonable to tidy it up though. (`preventDefault()` wouldn't guarantee that a notification is shown.)

index.html
martinthomson: Maybe you should double-check that math.

```
$ python
>>> 2**63-1
9223372036854775807
>>> 2**53-1
9007199254740991
```

(I don't like argument from precedent, as far as that goes.  Especially when it results in long strings of digits like this one.)

annevk: Interesting, I wonder why IDL goes beyond max safe integer. Also, I should have used the unsigned variant here so either way this is wrong: https://webidl.spec.whatwg.org/#idl-unsigned-long-long

index.html
```diff
@@ -164,6 +164,281 @@ <h2>
...
+          <p>
+            A [=declarative push message=] has the following members:
+          </p>
+          <dl>
+            <dt>
+              <code>web_push</code> (required)
+            </dt>
+            <dd>
+              <p>
+                An integer that must be 9001. Used to disambiguate a [=declarative push message=]
```
beverloo: As much as I appreciate 9001, we should also consider 8030? (Referring to the Web Push RFC.)  Or some number that reflects the collective number of grey hairs we have over this stuff, which is definitely over 9000 😄

index.html
```diff
@@ -164,6 +164,281 @@ <h2>
...
+                from other JSON documents.
+              </p>
+            </dd>
+            <dt>
+              <code>notification</code> (required)
```
beverloo, line 216: Is there any way in which this can simply match [`NotificationOptions`](https://notifications.spec.whatwg.org/#dictdef-notificationoptions) with a required title and URL? I guess this being JSON makes that difficult? Both the partial selection of available options and the duplication this introduces makes it fragile and more difficult to reason about.

index.html
```diff
@@ -164,6 +164,281 @@ <h2>
...
+                <dd>
+                  <p>
+                    A boolean. When true causes a <code>pushnotification</code> to be dispatched to
+                    a service worker (if any).
+                  </p>
+                </dd>
+              </dl>
+            </dd>
+            <dt>
+              <code>app_badge</code>
```
beverloo: There really are three features being added here:
1. Declarative push notifications (w/ the Web Notification API),
2. Declarative push notifications that also fire a JavaScript event,
3. Declarative app badge updating (w/ the Badging API).

I think it'd be helpful to think about them separately. Intuitively I'm in favour of (1) and (3), however I have some concerns about (2).

A large part of our interest in declarative notifications comes from device resource usage improvements that it enables in Web Push message delivery, and from time-to-feedback improvements when users activate a Web Notification.

If the design were to stick to features (1) and (3), it would remove the need to launch a renderer process in Chrome entirely, which brings major savings (& with enough time and determination we could even trim much of our browser process). The `mutable` property would remove that optimisation, and based on our experience I expect it to be adopted for purposes of (at least) sending analytics pings back to the server. There may be other ways to solve for that use case too.

@beidson describes this as:
> We also propose a more flexible event handling model that developers can opt into to transform a push notification. This new model still depends on Service Workers, but provides the privacy-preserving guarantee that there is always a fallback notification to be shown, so pushes cannot trigger silent background runtime.

Existing implementations already solve for this: Chrome shows a default notification on behalf of the website, and IIRC Firefox and Edge penalize and eventually remove the underlying push subscription. I don't understand what benefit this brings over existing functionality.

I propose scoping out this concept from the proposal as it'll need more discussion.

index.html
annevk: I had a discussion with @domenic in https://github.com/whatwg/infra/pull/642 which will necessitate even more "duplication". At which point I think it's okay that it's slightly differently organized when that makes more sense for the JSON format. I will try to add some explanatory text so it's less confusing.

index.html
annevk: I'm happy to go with 8030. That will outdate itself as well in due course, but it's arguably a more relevant irrelevant number.

index.html
annevk: @beverloo in case you missed this, I split out this discussion into #391. We consider the mutable field a very important part of the overall value proposition as without it web developers cannot uniformly adopt this more robust delivery mechanism.

index.html
annevk: I went with a subset of options for now, but the idea is to have full parity. Perhaps it's less confusing if I put all of it in v1 and we just maintain them in tandem.

index.html
```diff
@@ -1082,29 +1654,18 @@ <h2>
           <li>Set |b| to the result of <a data-lt="extract a byte sequence">extracting a byte
           sequence</a> from the "`data`" member of |eventInitDict|.
           </li>
-          <li>Set the `data` attribute of the event to a new <a>PushMessageData</a> instance with
-          `bytes` set to |b|.
+          <li>Set the `data` attribute of the event to a new {{PushMessageData}} instance whose
+          [=PushMessageData/bytes=] is |b|.
```
marcoscaceres, line 1658:
```suggestion
          {{PushMessageData/bytes}} is |b|.
```

index.html
annevk: Once https://github.com/whatwg/notifications/pull/227 lands this can be checked here instead of `showNotification()`. Does that seem reasonable? Anything else remaining?

index.html
```diff
@@ -164,6 +165,555 @@ <h2>
...
+                <dt>
+                  <code>silent</code>
+                </dt>
+                <dd>
+                  <p>
+                    A boolean.
+                  </p>
+                </dd>
+                <dt>
+                  <code>require_interaction</code>
```
saschanaz: Do we have a strong reason to go for snake_case? I find the case is the only deviation from NotificationOptions.

annevk: https://github.com/whatwg/infra/pull/642 goes into the high-level thinking behind this. https://w3ctag.github.io/design-principles/#casing-rules has the relevant guideline.

index.html
```diff
@@ -1127,73 +1676,262 @@ <h2>
...
+              </li>
+              <li>
+                <p>
+                  Let |declarativeResult| be the result of running the [=/declarative push message
+                  parser=] given |bytes|, |origin|, |baseURL|, and |fallbackTimestamp|.
+                </p>
+              </li>
+              <li>
+                <p>
+                  If |declarativeResult| is not failure:
```
saschanaz, line 1698: It's sorta surprising that this implicitly parses all push payloads as declarative push and then falls back, I wonder an explicit option on subscription would make more sense.

index.html
saschanaz: (If the casing was same the parsing step could just say "parse it as NotificationOptions" after extracting `title` first. And then updating notification options wouldn't require updating both specs.)

index.html
annevk: That would not allow for transitioning existing subscriptions. Being fully backwards compatible is part of the design goals, see the original issue.

index.html
annevk: Right, I did that initially and then based on feedback ended up here.

index.html
```diff
@@ -56,7 +56,8 @@
       };
     </script>
   </head>
-  <body data-cite="service-workers FILEAPI secure-contexts hr-time permissions ECMASCRIPT">
+  <body data-cite=
+  "service-workers FILEAPI secure-contexts hr-time permissions ECMASCRIPT NOTIFICATIONS BADGING">
```
saschanaz: Err, is this case sensitive or can we make things consistent?

index.html
saschanaz: Sorta understandable from infra point but I guess the callers would have to deal with this casing in that case, like:

```js
onpush = async (ev) => {
  // Err, our little cute browser doesn't support declarative web push,
  // so we are here for nice progressive web.
  const { notification } = await ev.data.json();
  
  // This will mostly work until you use requireInteraction.
  // await registration.showNotification(notification.title, notification);
  
  const notificationOptions = {
    ...notification,
    requireInteraction: notification.require_interaction, // 😅
  };
  await registration.showNotification(notification.title, notificationOptions);
};
```

I wonder it would make sense to have an exception here to prevent extra boilerplate and head-scratching?

index.html
```diff
@@ -164,6 +165,555 @@ <h2>
...
+                  <p>
+                    A boolean.
+                  </p>
+                </dd>
+                <dt>
+                  <code>data</code>
+                </dt>
+                <dd>
+                  <p>
+                    Any JSON value.
```
saschanaz, line 347: Do we know how many existing callers set binary data (arraybuffer/blob etc) here? One can still use base64 though (and Gecko serializes it to base64 anyway).

index.html
```diff
@@ -164,6 +165,555 @@ <h2>
...
+              </p>
+            </li>
+            <li>
+              <p data-cite="INFRA">
+                If <var>message</var> is not a [=/map=], then return failure.
+              </p>
+            </li>
+            <li>
+              <p>
+                If <var>message</var>["`web_push`"] does not [=map/exist=] or is not 8030, then
```
saschanaz, line 438: Is this kind of check common for web APIs handling JSON payload? Feels sorta random, I wonder what about using [JSON-LD](https://www.w3.org/TR/json-ld/) like `"@context": "https://www.w3.org/ns/web-push"` as [ActivityPub does](https://w3c.github.io/activitypub/#Overview).

saschanaz (commented):
Can we split `mutable` out of this PR for now?

index.html
annevk: I'm not familiar with ReSpec.

index.html
saschanaz: I'm told that there's "please no JSON-LD" issue in https://github.com/w3ctag/design-principles/issues/239

saschanaz: Taking this back, I don't sej any alternative, and looking random is not harming anything.

index.html
hober: @marcoscaceres, do you know?

index.html
annevk: I guess we can escalate, but I'm inclined to accept this inconvenience: https://github.com/w3ctag/design-principles/issues/554.

index.html
annevk: I don't know, but there should be a variety of workarounds indeed.

index.html
```diff
@@ -1127,73 +1676,262 @@ <h2>
...
+                          If |result| is not failure, then set |notificationShown| to |result|'s
+                          [=push event result/notification shown=] and |appBadgeSet| to |result|'s
+                          [=push event result/app badge set=].
+                        </p>
+                      </li>
+                    </ol>
+                  </li>
+                  <li>
+                    <p>
+                      If |notificationShown| is false, then run the [=notification show steps=]
```
saschanaz, line 1741: Noticed during an internal discussion; this would involve cookies as current implemented in Firefox, and clearing site data would mean the cookie jar would become empty, which may be significant if the image URL can only be fetched with proper cookie (imagine an image attachment from a DM). Should we accept such breakage? It seems like that should be somehow covered in the spec at least. (related: https://github.com/whatwg/notifications/issues/5)

index.html
dimitribouniol: I'll add that as a server library tracking this proposal, my [adoption documentation](https://github.com/mochidev/swift-webpush/blob/aa06736b6e40f4e8e47e7e4c85655b93832a0097/README.md?plain=1#L203-L219) looks similarly to @saschanaz's, and I'd similarly push for aligning the property names so they could be passed as-is to `showNotification()`, if only to easy the migration story the community would need to push for to get this adopted:

```js
self.addEventListener('push', function(event) {
    const data = event.data?.json() ?? {};
    event.waitUntil((async () => {
        const notification = data.notification ?? {}
        /// Try parsing the data, otherwise use fallback content. DO NOT skip sending the notification, as you must display one for every push message that is received or your subscription will be dropped.
        const title = notification.title ?? "Your App Name";
        const body = notification.body ?? "New Content Available!";
        
        await self.registration.showNotification(title, { 
            body,
            requireInteraction: notification.require_interaction ?? false,
            ...notification,
        });
    })());
});
```

annevk: > (If the casing was same the parsing step could just say "parse it as NotificationOptions" after extracting title first. And then updating notification options wouldn't require updating both specs.)

This would diverge too much from the standard JSON processing model so I don't think we should do that. But I think we can align on the casing here.

index.html
annevk: It's not entirely clear to me what you are suggesting here.

saschanaz: I'm saying that the image fetch step can no longer include cookies when site data is gone and the spec should say something about that.

annevk: Hmm. Isn't that just an artifact from how cookies work as well? Should every specification that potentially fetches something later also point out that cookies currently expire after a little over a year?

saschanaz: If that spec expects to fetch something without page visit for a long time, maybe? (AFAIK push event is the rare case doing that if not only)

saschanaz: (But good point, it's not really DWP specific but also affects the existing Push)

index.html
```diff
@@ -1127,73 +1680,262 @@ <h2>
...
+                      given |notification|.
+                    </p>
+                  </li>
+                  <li>
+                    <p>
+                      If |appBadgeSet| is false, then <a href=
+                      "https://github.com/w3c/badging/issues/111">w3c/badging #111</a>...
```
saschanaz: (This part is incomplete)

index.html
```diff
@@ -1127,73 +1680,262 @@ <h2>
...
+                </p>
+              </li>
+              <li>
+                <p>
+                  Let |fallbackTimestamp| be [=current coarsened wall time=].
+                </p>
+              </li>
+              <li>
+                <p>
+                  Let |declarativeResult| be the result of running the [=/declarative push message
```
saschanaz, line 1692: This happens after "If |registration| is not found, abort these steps." step above. Either #393 or this PR should fix that.

index.html
```diff
@@ -1127,73 +1680,262 @@ <h2>
...
+                  </li>
+                  <li>
+                    <p>
+                      Let |appBadgeSet| be false.
+                    </p>
+                  </li>
+                  <li>
+                    <p>
+                      If |declarativeResult|'s [=declarative push message parser result/mutable=]
+                      is true:
```
saschanaz, line 1721: Probably should check the existence of `registration` too?

index.html
```diff
@@ -1127,73 +1680,262 @@ <h2>
...
+          </li>
           <li>
             <p>
               <a>Fire a functional event</a> named "`push`" using <a>PushEvent</a> on
               |registration| with the following properties:
             </p>
             <dl>
               <dt>
-                `data`
+                {{PushEvent//data}}
```
saschanaz:
```suggestion
                {{PushEvent/data}}
```

saschanaz (commented):
Mostly fine with a few changes. We explicitly did not cover the app badge part of this PR though.

index.html
annevk: I would suggest we fix that in #393. And we merge this first.

index.html
annevk: Per above it's guaranteed to be non-null here, no?

index.html
annevk: Would you prefer splitting this out or leaving it as an open issue? I suppose I can create a new PR for this after we merge this. That might be cleanest.

saschanaz: Splitting this out would be nice, or the spec would look incomplete.

index.html
saschanaz: wfm.

saschanaz:
Force push breaks "Changes since your last review" feature 😵

index.html
annevk: Done. I think the main thing blocking this now are the Notifications API changes.

annevk:
I made https://github.com/w3c/push-api/pull/385/commits/7551a5f3aec532cdb4afea2be9561dff1077b4a7 and https://github.com/w3c/push-api/pull/385/commits/a64358c6a5cdb048e407017b055322eb88a27213 before I squashed (to create the copy PR for app badge). Both super tiny commits. Removal of app badge is a separate commit again.

saschanaz (approved):
You can check Mozilla on the interest list. (Or well, I did)

saschanaz:
Looking at https://github.com/WebKit/WebKit/pull/18087, it seems WebKit shipped "pushnotification" event instead of reusing existing "push" event, deviating from this PR? 🤔

(Edit: Or WebKit changed it to "push" but the remnant of "pushnotification" is still in the source?)

Filed https://bugs.webkit.org/show_bug.cgi?id=296772

saschanaz:
Also WebKit [considers `mutable` as a member of `notification` instead of a member of the top level push message](https://searchfox.org/wubkat/rev/c46918d1cba365017d1df98e3d14f04db42320bc/Source/WebCore/Modules/notifications/NotificationJSONParser.cpp#112-113,142-143).

Filed https://bugs.webkit.org/show_bug.cgi?id=296770.

annevk:
Thanks for finding those bugs, we'll get them fixed!

Yttuf (approved):

PR #388, opened 2024-08-26
Push subscriptions are not associated with service workers

### LINKED ISSUE #384: What is a push subscription associated with?
I was trying to make sense of `subscribe()` but it did something unexpected. It looks at a service worker registration's active worker and uses that to obtain a push subscription. However, very early on in this document we have this description:

> Each [push subscription](https://w3c.github.io/push-api/#dfn-push-subscription) is associated with a [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) and a [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) has at most one [push subscription](https://w3c.github.io/push-api/#dfn-push-subscription).

What gives?

It does seem like rejection when there's no active worker is intentional: https://github.com/w3c/push-api/pull/230.

It also seems to me that a push subscription is not stored with a service worker as you should be able to replace the service worker and retain the subscription.

There's also some weirdness around "having" a push subscription apparently being known information but "getting" the push subscription requiring a "request" (whatever that is).

annevk:
Fixes #384.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/388.html" title="Last updated on Aug 26, 2024, 3:52 PM UTC (8a4a87a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/388/019c701...8a4a87a.html" title="Last updated on Aug 26, 2024, 3:52 PM UTC (8a4a87a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -626,22 +626,18 @@
         global task=] on the [=networking task source=] using |global| to [=reject=] |promise| with
         an {{"InvalidStateError"}} {{DOMException}} and terminate these steps.
         </li>
-        <li>Let |sw| be |registration|'s [=service worker registration/active worker=].
-        </li>
         <li>Let |permission| be [=request permission to use=] "push".
         </li>
         <li>If |permission| is {{PermissionState/"denied"}}, [=queue a global task=] on the [=user
         interaction task source=] using |global| to [=reject=] |promise| with a
         {{"NotAllowedError"}} {{DOMException}} and terminate these steps.
         </li>
-        <li>If |sw| is already subscribed, run the following sub-steps:
+        <li>If |registration| has a <a>push subscription</a>:
           <ol>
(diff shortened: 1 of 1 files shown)


saschanaz (commented):
LGTM but maybe @asutherland want to double check?

That said I think Gecko still maps push with the SW scope instead of each registration: https://searchfox.org/mozilla-central/rev/aee7c3a0dbf33af0c4f6648f391db62b35895e50/dom/push/PushManager.cpp#506 (which we want to fix)

asutherland (approved):
Yeah, the coupling to ServiceWorker here seems like it was in error regardless of what we do as a result of https://github.com/w3c/push-api/pull/368.

Issue #205, opened 2016-07-22
Define behaviour of pushsubscriptionchange in the event of an error


benlast:
When I say "the spec", I'm referring to [https://w3c.github.io/push-api/#the-pushsubscriptionchange-event](https://w3c.github.io/push-api/#the-pushsubscriptionchange-event).

It's not defined that a `pushsubscriptionchange` event will be fired when the browser has network connectivity, and of course it's also possible that any given network request may fail for a variety of reasons.

In a `pushsubscriptionchange` event handler, it may be a common requirement for the service worker to update new subscription details to some application server. It's not clear how to handle the case of a failure in the network upload.

As a suggestion; If the network Promise is included in `event.waitUntil`, and the Promise rejects, the event may be re-sent to allow the service worker code to retry.

Otherwise, the event handler code must flag that a retry is required, and wait for some opportunity to retry the upload. However, it's not clear what opportunity it will get. Since the spec says that any old subscription is unsubscribed on completion of the `pushsubscriptionchange` event, no further `push` events may be expected, and there is no guarantee that there will be any other event (such as `install` or `activate`) fired.

benlast:
I see this is effectively included in Kit Cambridge's response at https://github.com/w3c/push-api/issues/132#issuecomment-225308850 so I'm closing this issue.

PR #393, opened 2024-11-19
Expose pushManager on Window

annevk:
This is part of the Declarative Web Push initiative (see #360) and mainly makes sense when that is supported, although could be independently supported in theory.

This makes window.pushManager work by making push subscriptions tied to a scope rather than a service worker registration. Most often push subscriptions remain 1:1 with service worker registrations, but the scope whose serialized path is "/" is treated specially from now on and can exist independently.

This obsoletes #368.

 * [x] Modified Web platform tests (link to pull request): N/A, except IDL coverage

Implementer support:
 
 * [ ] Chromium 
 * [x] Gecko
 * [x] WebKit

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/393.html" title="Last updated on Sep 8, 2025, 2:46 PM UTC (b0bfe7b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/393/f28f781...b0bfe7b.html" title="Last updated on Sep 8, 2025, 2:46 PM UTC (b0bfe7b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1166,7 +1166,6 @@
             <li>Set |scope| to the result of running the [=basic URL parser=] given "`/`" and
             |global|'s <a>associated <code>Document</code></a>'s [=Document/URL=].
             </li>
-            <!-- Should the scope be validated here or do we trust the permission story to take care of that? Hmm. -->
           </ol>
         </li>
         <li>Otherwise:
@@ -1307,7 +1306,7 @@
         <li>Let |registration| be null.
         </li>
         <li>If [=this=]'s [=PushManager/service worker registration=] is null, then set
-        |windowScope| to the result of running the [=basic URL parser=] given "`/`" and [=this=]'s
+        |windowScope| to the result of running the [=basic URL parser=] given "`/`" and |global|'s
         <a>associated <code>Document</code></a>'s [=Document/URL=].
         </li>
         <li>Otherwise:
--- a/index.html
+++ b/index.html
@@ -1097,7 +1097,12 @@
         ServiceWorkerRegistration includes PushManagerAttribute;
       </pre>
       <p>
-        The <dfn>pushManager</dfn> attribute exposes a {{PushManager}}.
+        {{Window}} and {{ServiceWorkerRegistration}} objects have an associated {{PushManager}}
+        object.
+      </p>
+      <p>
+        The <dfn>pushManager</dfn> getter steps are to return [=this=]'s associated {{PushManager}}
+        object.
       </p>
       <p>
         On a {{Window}} object the {{PushManager}}'s [=PushManager/service worker registration=] is
@@ -1184,10 +1189,14 @@
             </p><!-- This doesn't really seem acceptable this day and age. -->
           </aside>
           <ol>
+            <li>If |scope| is failure or is a [=/URL=] whose [=url/scheme=] is not "`https`", then
+            [=queue a global task=] on the [=networking task source=] using |global| to [=reject=]
+            |promise| with a {{"NotAllowedError"}} {{DOMException}}.
+            </li>
             <li>If the |options| argument has a {{PushSubscriptionOptionsInit/userVisibleOnly}}
             value set to `false` and the user agent requires it to be `true`, [=queue a global
-            task=] on the [=networking task source=] using |global| to [=reject=] |promise|
-            {{"NotAllowedError"}} {{DOMException}}
+            task=] on the [=networking task source=] using |global| to [=reject=] |promise| with a
+            {{"NotAllowedError"}} {{DOMException}}.
             </li>
             <li>If the |options| argument does not include a non-null value for the
             {{PushSubscriptionOptionsInit/applicationServerKey}} member, and the <a>push
--- a/index.html
+++ b/index.html
@@ -699,9 +699,19 @@
         </h2>
         <p>
           A <dfn>push subscription</dfn> is a message delivery context established between the
-          <a>user agent</a> and the <a>push service</a> on behalf of a web application. Each
-          <a>push subscription</a> is associated with a <a>service worker registration</a> and a
-          <a>service worker registration</a> has at most one <a>push subscription</a>.
+          <a>user agent</a> and the <a>push service</a> on behalf of a web application.
+        </p>
+        <p>
+          A [=push subscription=] has an associated <dfn data-dfn-for=
+          "push subscription">scope</dfn>, which is a [=/URL=].
+        </p>
+        <p>
+          A [=push subscription=] is considered to have a <dfn data-dfn-for=
+          "push subscription">window-accessible scope</dfn> when its [=push subscription/scope=] is
+          a [=/list=] of [=list/size=] 1 and [=push subscription/scope=][0] is the empty string.
+        </p>
+        <p class="note">
+          I.e., the [=url/path=] component of the [=/URL=] serializes as "`/`".
         </p>
         <p>
           A <a>push subscription</a> has an associated <dfn>push endpoint</dfn>. It MUST be the
@@ -721,12 +731,9 @@
           creating the <a>push subscription</a>.
         </p>
         <p>
-          If the <a>user agent</a> has to change the keys for any reason, it MUST <a>fire the
-          "`pushsubscriptionchange`" event</a> with the <a>service worker registration</a>
-          associated with the <a>push subscription</a> as |registration|, a {{PushSubscription}}
-          instance representing the <a>push subscription</a> having the old keys as
-          |oldSubscription| and a {{PushSubscription}} instance representing the <a>push
-          subscription</a> having the new keys as |newSubscription|.
+          If the <a>user agent</a> has to change the keys of a [=push subscription=] for any reason
+          and the [=push subscription=]'s [=associated service worker registration=] is non-null,
+          it MUST [=refresh=] the [=push subscription=].
         </p>
         <p>
           To <dfn>create a push subscription</dfn>, given an {{PushSubscriptionOptionsInit}}
@@ -771,17 +778,41 @@
         </ol>
         <section>
           <h2>
-            Subscription Refreshes
+            Relationship to service worker registrations
+          </h2>
+          <p>
+            A [=push subscription=]'s <dfn>associated service worker registration</dfn> is the
+            [=service worker registration=] whose [=service worker registration/scope URL=]
+            [=URL/equals=] the [=push subscription=]'s [=push subscription/scope=], if any;
+            otherwise null.
+          </p>
+          <p class="note">
+            A [=push subscription=]'s [=associated service worker registration=] can only be null
+            when it has a [=push subscription/window-accessible scope=].
+          </p>
+          <p>
+            And vice versa, a [=service worker registration=]'s <dfn>associated push
+            subscription</dfn> is the [=push subscription=] whose [=push subscription/scope=]
+            [=URL/equals=] the [=service worker registration=]'s [=service worker
+            registration/scope URL=], if any; otherwise null.
+          </p>
+        </section>
+        <section>
+          <h2>
+            Subscription refreshes
           </h2>
           <p>
             A <a>user agent</a> or <a>push service</a> MAY choose to <dfn>refresh</dfn> a <a>push
-            subscription</a> at any time, for example because it has reached a certain age.
+            subscription</a> whose [=associated service worker registration=] is non-null at any
+            time, for example because it has reached a certain age.
           </p>
           <p>
             When this happens, the <a>user agent</a> MUST run the steps to <a>create a push
             subscription</a> given the <a>PushSubscriptionOptions</a> that were provided for
-            creating the current <a>push subscription</a>. The new <a>push subscription</a> MUST
-            have a key pair that's different from the original subscription.
+            creating the current <a>push subscription</a>, and set the new [=push subscription=]'s
+            [=push subscription/scope=] to the original subscription's [=push subscription/scope=].
+            The new <a>push subscription</a> MUST have a key pair that's different from the
+            original subscription.
           </p>
           <p>
             When successful, <a>user agent</a> then MUST <a>fire the "`pushsubscriptionchange`"
@@ -808,7 +839,7 @@
         </section>
         <section>
           <h2>
-            Subscription Deactivation
+            Subscription deactivation
           </h2>
           <p>
             When a <a>push subscription</a> is <dfn data-lt="deactivate">deactivated</dfn>, both
@@ -817,13 +848,13 @@
             delivered.
           </p>
           <p>
-            A <a>push subscription</a> is <a>deactivated</a> when its associated <a>service worker
-            registration</a> is unregistered, though a <a>push subscription</a> MAY be
-            <a>deactivated</a> earlier.
+            A <a>push subscription</a> without a [=push subscription/window-accessible scope=] is
+            <a>deactivated</a> when its associated <a>service worker registration</a> is
+            unregistered, though a <a>push subscription</a> MAY be <a>deactivated</a> earlier.
           </p>
           <p class="note">
-            A <a>push subscription</a> is removed when <a>service worker registration</a> is
-            cleared.
+            A <a>push subscription</a> without a [=push subscription/window-accessible scope=] is
+            removed when <a>service worker registration</a> is cleared.
           </p>
         </section>
       </section>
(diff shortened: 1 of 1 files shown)


saschanaz:
@asutherland previously had a strong opinion that notifications should be paired with registration instead of scope in #205, so adding him as a reviewer.

(Although now I see https://github.com/w3c/push-api/pull/368#issuecomment-2318273604)

index.html
```diff
@@ -192,12 +202,9 @@ <h2>
...
         <p>
-          If the <a>user agent</a> has to change the keys for any reason, it MUST <a>fire the
-          "`pushsubscriptionchange`" event</a> with the <a>service worker registration</a>
-          associated with the <a>push subscription</a> as |registration|, a {{PushSubscription}}
-          instance representing the <a>push subscription</a> having the old keys as
-          |oldSubscription| and a {{PushSubscription}} instance representing the <a>push
-          subscription</a> having the new keys as |newSubscription|.
+          If the <a>user agent</a> has to change the keys of a [=push subscription=] for any reason
+          and the [=push subscription=]'s [=associated service worker registration=] is non-null,
+          it MUST [=refresh=] the [=push subscription=].
```
saschanaz, line 737: (Reviewer note: this change is an editorial refactoring to reuse the term "refresh")

index.html
```diff
@@ -242,17 +249,41 @@ <h2>
...
+        </section>
+        <section>
+          <h2>
+            Subscription refreshes
           </h2>
           <p>
             A <a>user agent</a> or <a>push service</a> MAY choose to <dfn>refresh</dfn> a <a>push
-            subscription</a> at any time, for example because it has reached a certain age.
+            subscription</a> whose [=associated service worker registration=] is non-null at any
+            time, for example because it has reached a certain age.
```
saschanaz, line 808: Does this mean a subscription without an SWR is never supposed to be refreshed? I think the Firefox push server will keep its logic to auto-clear subscriptions (we do that when an inactive subscription is being spammed and gets a huge amount of message queue). cc @jrconlin

index.html
```diff
@@ -520,23 +551,23 @@ <h3>
...
-        The Service Worker specification defines a {{ServiceWorkerRegistration}} interface
-        [[SERVICE-WORKERS]], which this specification extends.
+        This specifications extends the {{Window}} and {{ServiceWorkerRegistration}} objects
+        through the {{PushManagerAttribute}} mixin. [[HTML]] [[SERVICE-WORKERS]]
       </p>
-      <pre class="idl" data-cite="service-workers">
+      <pre class="idl" data-cite="html service-workers">
         [SecureContext]
-        partial interface ServiceWorkerRegistration {
+        interface mixin PushManagerAttribute {
```
saschanaz, line 1093: Sorta weird name when it's a mixin and not an attribute. Bikeshedding: PushManagerBase (kinda following WebGL), PushManagerCommon (kinda following TextDecoder/Encoder), PushManageable (kinda following Animatable), PushManagerSource (kinda following FontFaceSource), PushManagerMixin (kinda following ARIAMixin and PaintTimingMixin), etc etc...

Not strongly against the current name though, especially it's just a spec-only name.

index.html
```diff
@@ -578,7 +609,23 @@ <h3>
...
         <li>Let |global| be [=this=]' [=relevant global object=].
         </li>
-        <li>Return |promise| and continue [=in parallel=].
+        <li>Let |scope| be null.
+        </li>
+        <li>Let |registration| be null.
+        </li>
+        <li>If [=this=] is a {{Window}} object, then set |scope| to the result of running the
+        [=basic URL parser=] given "`/`" and |global|'s <a>associated <code>Document</code></a>'s
+        [=Document/URL=].
```
saschanaz: This would fail for `about:blank`, but I guess that's fine as this is restricted to SecureContext.

index.html
```diff
@@ -588,80 +635,106 @@ <h3>
...
-        </li>
-        <li>If the |options| argument includes a non-null value for the
-        {{PushSubscriptionOptions/applicationServerKey}} attribute, run the following sub-steps:
           <ol>
-            <li>If |options|'s {{PushSubscriptionOptionsInit/applicationServerKey}} is a
-            {{DOMString}}, set its value to an {{ArrayBuffer}} containing the sequence of octets
-            that result from decoding |options|'s
-            {{PushSubscriptionOptionsInit/applicationServerKey}} using the base64url encoding
-            [[RFC7515]].
+            <li>If the |options| argument has a {{PushSubscriptionOptionsInit/userVisibleOnly}}
```
saschanaz, line 1195: (Reviewer note: nothing is changed here except indentation and rewrapping, following the "Run these steps in parallel" change and until the scope check below. It would be super nice if there was a smarter diff viewer for human language for this kind of changes.)

index.html
```diff
@@ -588,80 +635,106 @@ <h3>
...
-            <li>When the request has been completed, [=queue a global task=] on the [=networking
-            task source=] using |global| to [=resolve=] |promise| with |subscription| and terminate
-            these steps.
+            <li>If |permission| is {{PermissionState/"denied"}}, [=queue a global task=] on the
+            [=user interaction task source=] using |global| to [=reject=] |promise| with a
+            {{"NotAllowedError"}} {{DOMException}} and terminate these steps.
+            </li>
+            <li>If |subscription| is non-null:
+              <ol>
+                <li>If there is an error with |subscription|, then [=queue a global task=] on the
```
saschanaz, line 1261: This sentence now sounds like there's an error field on subscriptions, but given the previous sentence was "Let subscription be the result of obtaining registration's push subscription. If there is an error, ..." it was more about an error while obtaining the subscription.

I think either we should throw earlier in the subscription obtaining steps or just remove this step, as the obtaining steps are now clearer and infallible (except the existing null check)?

index.html
```diff
@@ -588,80 +635,106 @@ <h3>
...
+                |subscription|, then [=queue a global task=] on the [=networking task source=]
+                using |global| to [=reject=] |promise| with an {{"InvalidStateError"}}
+                {{DOMException}} and terminate these steps.
+                </li>
+                <li>[=Queue a global task=] on the [=networking task source=] using |global| to
+                [=resolve=] |promise| with |subscription| and terminate these steps.
+                </li>
+              </ol>
+            </li>
+            <li>[=/Assert=]: |subscription| is null and |scope| is a [=/URL=].
```
saschanaz, line 1279: (The null check assertion here is a bit silly, we're right below the null check above... 😛)

index.html
```diff
@@ -671,17 +744,54 @@ <h3>
...
+                <li>[=/Assert=]: |registration| is non-null.
+                </li>
+                <li>If |registration|'s [=associated push subscription=] is non-null, then set
+                |subscription| to |registration|'s [=associated push subscription=].
+                </li>
+              </ol>
+            </li>
+            <li>If |subscription| is null, then resolve |promise| with null.
+            </li>
+            <li>If there is an error with |subscription|, reject |promise| with a {{DOMException}}
```
saschanaz, line 1343: (Same about error, might be nicer if we throw in the obtaining steps or just skip this. It's unclear what a caller should do if this function throws, so maybe implementations should just pretend that no subscription exists if it's corrupted or something. If there's a good reason to throw error, it would be nice to have a note about that.)

index.html
```diff
@@ -288,13 +319,13 @@ <h2>
             delivered.
           </p>
           <p>
-            A <a>push subscription</a> is <a>deactivated</a> when its associated <a>service worker
-            registration</a> is unregistered, though a <a>push subscription</a> MAY be
-            <a>deactivated</a> earlier.
+            A <a>push subscription</a> without a [=push subscription/window-accessible scope=] is
+            <a>deactivated</a> when its associated <a>service worker registration</a> is
+            unregistered, though a <a>push subscription</a> MAY be <a>deactivated</a> earlier.
```
saschanaz, line 854: Should we provide a way to migrate to window-accessible scope to make an existing non-root scoped subscription be independent from service worker, probably as a followup?

index.html
```diff
@@ -588,80 +635,106 @@ <h3>
...
+                </li>
+                <li>If |registration|'s [=service worker registration/active worker=] is null,
+                [=queue a global task=] on the [=networking task source=] using |global| to
+                [=reject=] |promise| with an {{"InvalidStateError"}} {{DOMException}} and terminate
+                these steps.
+                </li>
+                <li>If |registration|'s [=associated push subscription=] is non-null, then set
+                |subscription| to |registration|'s [=associated push subscription=].
+                </li>
+                <li>Set |scope| to |registration|'s [=service worker registration/scope URL=].
```
saschanaz, line 1249: If the scope URL is the root path then this is also accessible from window, right? I assume that's the intention.

saschanaz (commented):
My first round of review. Looks mostly fine but with some questions.

saschanaz:
Neither this PR nor #385 changes the step 2 in section 10.4 Recieving a Push Message:

> 1. Let registration be the [service worker registration](https://www.w3.org/TR/service-workers/#dfn-service-worker-registration) corresponding to the [push message](https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/393.html#dfn-push-message).
> 2. If registration is not found, abort these steps.

Without changing that I think the window-subscribed push messages would be ignored.

index.html
```diff
@@ -170,9 +170,19 @@ <h2>
...
+          <a>user agent</a> and the <a>push service</a> on behalf of a web application.
+        </p>
+        <p>
+          A [=push subscription=] has an associated <dfn for="push subscription">scope</dfn>, which
+          is a [=/URL=].
+        </p>
+        <p>
+          A [=push subscription=] is considered to have a <dfn for=
+          "push subscription">window-accessible scope</dfn> when its [=push subscription/scope=] is
+          a [=/list=] of [=list/size=] 1 and [=push subscription/scope=][0] is the empty string.
```
saschanaz: Here the scope seems to be a list of string, but right above it's defined as a URL struct, which looks like a conflict.

annevk: Indeed, this should look at the scope's path instead of scope.

index.html
annevk: I think that's already possible as once you have permission it's quite easy to get a new subscription in, but maybe that varies across implementations.

index.html
annevk: It's a mixin that holds an attribute. Anyway, happy to change to whatever.

index.html
annevk: I think you're correct that there might be some failure scenarios here this needs to account for.

index.html
saschanaz: That should be sufficient for Firefox at least, yup.

index.html
annevk: Well, the above set of steps is somewhat long and this is not an otherwise branch..

index.html
annevk: I think we should keep this as this is not something I was aiming to change in this PR. I'm not really sure how to do it earlier though without getting very verbose.

index.html
annevk: Clearing seems fine, but refreshing indeed doesn't work. See #401.

index.html
annevk: Given the lack of follow-up, was this okay after all?

index.html
annevk: That's right.

index.html
saschanaz: It's a mixin so this can change anytime, not blocking on this.

index.html
annevk: This is just trying to preserve existing text.

saschanaz:
A significant portion of this PR can be part of an editorial PR. I'll file a new one for them and merge it myself, for easier review here.

index.html
```diff
@@ -699,9 +699,20 @@ <h2>
...
+          "push subscription">scope</dfn>, which is a [=/URL=].
+        </p>
+        <p>
+          A [=push subscription=] is considered to have a <dfn data-dfn-for=
+          "push subscription">window-accessible scope</dfn> when its [=push subscription/scope=]'s
+          [=url/path=] is a [=/list=] of [=list/size=] 1 and [=push subscription/scope=]'s
+          [=url/path=][0] is the empty string.
+        </p>
+        <p class="note">
+          I.e., the [=url/path=] component of the [=/URL=] serializes as "`/`".
```
saschanaz, line 715: Can we just say this instead of checking the raw value? This is more readable and implementations can freely do the other way for optimization.

index.html
```diff
@@ -721,8 +732,9 @@ <h2>
           creating the <a>push subscription</a>.
         </p>
         <p>
-          If the [=user agent=] has to change the keys for any reason, it MUST [=refresh=] the
-          [=push subscription=].
+          If the <a>user agent</a> has to change the keys of a [=push subscription=] for any reason
+          and the [=push subscription=]'s [=associated service worker registration=] is non-null,
+          it MUST [=refresh=] the [=push subscription=].
```
saschanaz, line 737: (Might want to mention #401 in a note)

index.html
```diff
@@ -1111,13 +1186,17 @@ <h3>
               {{PushSubscriptionOptionsInit/userVisibleOnly}} is allowed after validating the
               {{PushSubscriptionOptionsInit/applicationServerKey}}, and vice versa). However, we
               don't believe this affects interoperability of implementations or web applications.
-            </p>
+            </p><!-- This doesn't really seem acceptable this day and age. -->
           </aside>
           <ol>
+            <li>If |scope| is failure or is a [=/URL=] whose [=url/scheme=] is not "`https`", then
```
saschanaz, line 1191: I wonder we can reject earlier, but maybe it's more consistent to do it here. But then I think the scope getting step can be here rather than above.

index.html
```diff
@@ -1103,6 +1157,27 @@ <h3>
...
+        <li>Let |scope| be null.
+        </li>
+        <li>Let |registration| be null.
+        </li>
+        <li>If [=this=]'s [=PushManager/service worker registration=] is null:
+          <ol>
+            <li>Set |scope| to the result of running the [=basic URL parser=] given "`/`" and
+            |global|'s <a>associated <code>Document</code></a>'s [=Document/URL=].
+            </li>
+            <!-- Should the scope be validated here or do we trust the permission story to take care of that? Hmm. -->
```
saschanaz: Can't think of a specific concern, can you?

index.html
saschanaz: Because throwing it earlier changes the order? Maybe that's a valid concern, although I think the order already is not compatible between implementations and we'll have to modify that part.

Maybe we can move the ["partial database corruption" part in `PushSubscriptionChangeEventInit`](https://w3c.github.io/push-api/#dom-pushsubscriptionchangeeventinit-oldsubscription) into PushSubscription to define what this is about, because with this change it's less clear what this error is about. (to be fair it was already unclear though.)

saschanaz (approved):
I don't see a blocker, just a few editorial things. I'm fine with dealing with them in separate issues.

saschanaz:
Section 4 about security consideration has this:

> When a service worker registration is unregistered, any associated push subscription MUST be deactivated.

We probably want to change that. (Either remove that or make it non-normative? Not sure why it should be defined again in a security section.)

Edit: Maybe the Subscription deactivation section should say MUST for deactivation after unregistration.

index.html
```diff
@@ -1201,17 +1300,51 @@ <h3>
...
-        <li>Run these steps [=in parallel=]:
+        <li>Let |global| be [=this=]'s [=relevant global object=].
+        </li>
+        <li>Let |windowScope| be null.
+        </li>
+        <li>Let |registration| be null.
+        </li>
+        <li>If [=this=]'s [=PushManager/service worker registration=] is null, then set
+        |windowScope| to the result of running the [=basic URL parser=] given "`/`" and [=this=]'s
+        <a>associated <code>Document</code></a>'s [=Document/URL=].
```
saschanaz: Should it use `global` instead of `this`?

index.html
```diff
@@ -1644,20 +1777,34 @@ <h2>
...
         </p>
         <ol>
-          <li>Let |registration| be the <a>service worker registration</a> corresponding to the <a>
-            push message</a>.
-          </li>
-          <li>If |registration| is not found, abort these steps.
+          <li>
+            <p>
+              Let |subscription| be the active <a>push subscription</a> corresponding to the
+              <a>push message</a>.
```
saschanaz, line 1782: It seems there's no spec described way to get a subscription from a push message (nor get a registration either).

index.html
annevk: It would be slightly different if you somehow ended up with a URL object that violates the contract. That's a bug of course, but I'd rather do the correct thing here and not hide the gory details.

index.html
annevk: Right, it's more consistent to do it here. And no, we don't want to grab state while in parallel.

index.html
annevk: Now that I added the failure check below this note is indeed no longer needed.

index.html
annevk: I'm happy to file a follow-up issue for this and then people more experienced with these algorithms can figure it out.

index.html
annevk: Yeah, I can file another issue on that. This seems okay since it was already vague.

annevk: https://github.com/w3c/push-api/issues/408

index.html
annevk: https://github.com/w3c/push-api/issues/408

saschanaz: #408 is not really about error?

annevk: It's not?

> And as the specification hints that this process could fail, we should clarify that at the same time.

saschanaz: I don't see how it is relevant to the relationship between pushmessage and subscription...

annevk: Can you file an issue then, since I'm honestly not sure what you're trying to say? Ultimately we're trying to obtain the subscription of a subscription, perhaps via a registration (or now a scope), and that process can fail. None of that is really spelled out at the moment, but it would be good if it were.

saschanaz: > obtain the subscription of a subscription

object of a subscription?

The process can fail in case the local PushDB somehow is broken and thus relevant data e.g. the endpoint URL and public key can't be accessed (in case of Firefox), but push messages are from push server and thus never stored in a DB.

From #408's title one cannot easily link the issue and the DB breakage, as that's nothing to do with push messages - it can break without any message.

annevk: Sure, but the title alone is not the whole issue...

saschanaz: For now I changed the title, if that's good for you we can resolve this one.

PR #402, opened 2025-07-30
Add app badge to Declarative Web Push

annevk:
To land this https://github.com/w3c/badging/issues/111 needs to be completed. I'd like some help with that. @marcoscaceres maybe?

---

Implementation support:
 
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)
 * [x] WebKit

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/402.html" title="Last updated on Sep 15, 2025, 2:21 AM UTC (cfc1388)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/push-api/402/54d5cb1...cfc1388.html" title="Last updated on Sep 15, 2025, 2:21 AM UTC (cfc1388)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1779,11 +1779,26 @@
                       given |notification|.
                     </p>
                   </li>
-                  <li>
+                  <li data-cite="appmanifest">
                     <p>
-                      If |appBadgeSet| is false, then <a href=
-                      "https://github.com/w3c/badging/issues/111">w3c/badging #111</a>...
+                      If |appBadgeSet| is false and the web application associated with the <a>push
+                      subscription</a> is an [=installed web application=]:
                     </p>
+                    <ol>
+                      <li>
+                        <p>
+                          Let |identity| be the web application's [=installed web
+                          application/identity=] [=/URL=].
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          [=Set the application badge for installed web application=] given
+                          |identity| and |declarativeResult|'s [=declarative push message parser
+                          result/app badge=].
+                        </p>
+                      </li>
+                    </ol>
                   </li>
                   <li>
                     <p>
--- a/index.html
+++ b/index.html
@@ -393,13 +393,25 @@
               </dl>
             </dd>
             <dt>
+              <code>app_badge</code>
+            </dt>
+            <dd>
+              <p>
+                A [=/64-bit unsigned integer=].
+              </p>
+              <p class="note">
+                Platform conventions are likely to impose a lower limit with regards to what is
+                displayed to the end user. [[BADGING]]
+              </p>
+            </dd>
+            <dt>
               <code>mutable</code>
             </dt>
             <dd>
               <p>
                 A boolean. When true causes a <code>push</code> event to be dispatched to a service
-                worker (if any) containing the {{Notification}} object described by the
-                <a>declarative push message</a>.
+                worker (if any) containing the {{Notification}} object and app badge number (if
+                any) described by the <a>declarative push message</a>.
               </p>
             </dd>
           </dl>
@@ -411,7 +423,8 @@
           <p>
             A <dfn>declarative push message parser result</dfn> is a [=/tuple=] consisting of a
             <dfn data-dfn-for="declarative push message parser result">notification</dfn> (a
-            [=/notification=]) and a <dfn data-dfn-for=
+            [=/notification=]), an <dfn data-dfn-for="declarative push message parser result">app
+            badge</dfn> (null or an integer), and a <dfn data-dfn-for=
             "declarative push message parser result">mutable</dfn> (a boolean).
           </p>
           <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1742,6 +1779,27 @@ <h2>
...
+                    <ol>
+                      <li>
+                        <p>
+                          Let |identity| be the web application's [=installed web
+                          application/identity=] [=/URL=].
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          [=Set the application badge for installed web application=] given
```
marcoscaceres, line 1796: See https://github.com/w3c/badging/pull/116/

marcoscaceres:
@annevk, to kick this off:

 * I sent a PR to add "[set the application badge for installed web application](https://github.com/w3c/badging/pull/116/)" to the badging spec (no platform object needed)
 * In a hand-wavy way 👐, the spec checks if the "web application" is an "installed web application", and gets its "identifier" (a URL).
 * Once it has the identifier, it uses that to set the badge number for the installed web application.  

happy to refine from here... but it's broadly what needs to happen, right?

index.html
```diff
@@ -392,14 +392,26 @@ <h3>
                 </dd>
               </dl>
             </dd>
+            <dt>
+              <code>app_badge</code>
+            </dt>
+            <dd>
+              <p>
+                A [=/64-bit unsigned integer=].
```
marcoscaceres: Just noting that not all platforms support setting a number: for instance, Android allows setting the badge to "flag" (which just shows up as a dot). Some macOS apps do this too (e.g., Slack). 

For backwards compat, we might need to default change the badging API to support an enum to explicitly set a `"flag"` as the default value, so that:

 * setAppBadge() - "flag"
 * setAppBadge(0) - clear
 * setAppBadge(N) - number.

annevk: So you're saying that even platforms that do render a number, can also render just the badge without the number? In that case we'd indeed need to support that separate state. We could perhaps allow passing undefined or null for that? Bit finicky, I'll admit.

index.html
```diff
@@ -1830,21 +1901,28 @@ <h2>
...
+                  successfully invoked=].
                 </p>
               </li>
               <li>
                 <p>
-                  Otherwise, set |notificationResult| to |registration|'s [=service worker
-                  registration/has shownotification() been successfully invoked|has
-                  `showNotification()` been successfully invoked=].
+                  Set |appBadgeResult| to true if {{NavigatorBadge/setAppBadge()}} has been
+                  invoked; otherwise false.
```
annevk, line 1918: @marcoscaceres we need a hook for this similar to "has `showNotification()` been successfully invoked". People were not happy with this being defined in a hand-wavy way.

index.html
```diff
@@ -1742,6 +1779,27 @@ <h2>
                       given |notification|.
                     </p>
                   </li>
+                  <li data-cite="appmanifest">
+                    <p>
+                      If |appBadgeSet| is false and the web application associated with the <a>push
+                      subscription</a> is an [=installed web application=]:
```
annevk, line 1785: We should use |subscription| here, I think? Maybe

> If appBadgeSet is false and |subscription| has an associated installed web application:

annevk: Though it would be nice to also support this outside of installed web applications. But I suppose that's a separate change.

index.html
marcoscaceres: > So you're saying that even platforms that do render a number, can also render just the badge without the number?

Correct. Android in particular supports both modes ("[App Icon can Show Badges with Numbers or Dot-style Badges](https://www.samsung.com/hk_en/support/mobile-devices/android-o-os-app-icon-can-show-badges-with-numbers-or-dot-style-badges/)"). Even though it's user controlled, a web developer can still just set a "dot" (i.e., what in the Badging spec is called a "flag"). 

> Bit finicky, I'll admit.

Yeah :( hopefully we can do it consistently across both specs. 

Possibly the most compatible option would be to change `app_badge` to support either a `boolean` or a `64-bit unsigned integer`: 

1. If app_badge is missing: do nothing.
2. If app_badge === true: then "flag.
3. Else if app_badge === false: "clear" (treat at 0).
4. Else if typeof app_badge == "number", set number. 

We can distinguish between the types, so could potentially work.

index.html
marcoscaceres: Setting a badge is pretty tightly bound to installation at the moment, as that's how it's been implemented by everyone. We could change this if this changes.

index.html
marcoscaceres: IIUC, in the case of notifications, it emulates calling `showNotification()`... 

Instead, for the Badging case, can we reuse "set the application badge for an installed web application" and I can get that to return a boolean if it worked?

index.html
annevk: That shape sounds good. What happens for platforms that don't support "flag" though?

index.html
annevk: No, this is specifically to catch the case where the developer used `setAppBadge()` (and maybe even `clearAppBadge()`, come to think of it) and so we don't want to use the declarative app badge result.

See https://notifications.spec.whatwg.org/#service-worker-registration-has-shownotification-been-successfully-invoked and how `showNotification()` ends up setting it to true. (We set it to false here first before dispatching the event to the service worker.) We need the same for app badge.

index.html
marcoscaceres: Described in https://github.com/w3c/badging/pull/123

marcoscaceres:
```suggestion
                A boolean or a [=/64-bit unsigned integer=].
```

index.html
marcoscaceres: ah, ok... I'll have another look.

